PR #26, opened 2026-02-27
feat: add LendaSwap provider

Kukks:
## LendaSwap provider (v0.2.34) on the multi-provider swap framework

Adds [LendaSwap](https://github.com/satoraHQ/lendaswap-sdk) as a second `ISwapProvider` alongside Boltz, extending the swap router to cover Ark↔EVM (Ethereum / Polygon / Arbitrum), Ark↔BTC-onchain, and Lightning↔EVM routes that Boltz doesn't cover.

### What's in the branch

- **HTTP client** (`NArk.Swaps/LendaSwap/Client/`): typed `LendaSwapClient` with `Quotes` / `Swaps` / `Status` partials covering the three live endpoints (`POST /api/vtxo-swap`, `GET /api/vtxo-swap/{id}`, `GET /api/vtxo-swap/estimate`).
- **Provider** (`LendaSwapProvider`): implements `ISwapProvider`, routes Ark↔EVM/BTC/LN, polls active swaps every 30s, raises `SwapStatusChangedEvent` on transitions, maps LendaSwap status strings to `ArkSwapStatus`.
- **Routing extensions**: `SwapAsset.Erc20(chain, contractAddress)` factory + `SwapNetwork.EvmEthereum`/`EvmPolygon`/`EvmArbitrum` enum values (additive — master's `BoltzSwapProvider` is unaffected).
- **DI**: `AddLendaSwapProvider(Action<LendaSwapOptions>?)` extension on `SwapServiceCollectionExtensions`.
- **Tests**: unit tests for `LendaSwapClient` (mocked HTTP) + `SwapRouting`; E2E tests with server-availability gating so they're skipped when the LendaSwap server is offline.

### Rebased onto current master (was 33 commits behind)

The branch was originally proposed before the multi-provider architecture (PR #79) landed. The foundational pieces — `SwapNetwork`/`SwapAsset`/`ISwapProvider`/the `BoltzSwapProvider` + `SwapsManagementService` refactor — were already extracted and merged. This rebase keeps **only the LendaSwap-specific commits** (8 commits) and reconciles them with master's modern API:

- `LendaSwapProvider.StartAsync` updated to the new `(CancellationToken)` signature.
- Removed dead `CreateSwapAsync` / `RefundSwapAsync` overrides (the interface drops them).
- Test mocks updated from `IChainTimeProvider` → `IBitcoinBlockchain` (the unified blockchain abstraction master adopted via #53).

### Brought up to LendaSwap SDK v0.2.34

Three upstream changes since v0.2.31 (when this branch was last touched against) are reflected:

1. **`extra_fees` (v0.2.33)** — optional per-swap surcharge in basis points (0..max_extra_fee_bps configured on the matching developer key on the LendaSwap side). Added as a nullable `int? ExtraFees` on **all 10 `Create*Request` DTOs** + as an optional `extraFees:` parameter on `LendaSwapClient.GetQuoteAsync`.
2. **`referral_code` (v0.2.34)** — wire field renamed from `reflink_code`. All request DTOs renamed `ReflinkCode` → `ReferralCode` (with the `referral_code` JSON wire name). `LendaSwapOptions.ReferralCode` is the canonical client-level default; `ReflinkCode` retained as `[Obsolete]` alias mirroring the ts-sdk's `withOrgCode` deprecation-not-removal policy.
3. **Detailed API status (v0.2.32)** — the existing status response already exposes the per-swap status string we consume; no additional client surface required for the integration.

The PR description used to cite v0.2.20 — that tag is gone from upstream history (releases were rewritten); current is v0.2.34, and the branch is now pinned to it.

### Test plan

- [x] `dotnet build` clean across `NArk.Swaps`, `NArk.Tests`, `NArk.Tests.End2End`.
- [x] **437/437 unit tests pass** locally (`dotnet test NArk.Tests`).
- [ ] CI green on the rebased commit (`1013cde`).
- [ ] E2E green when the LendaSwap test server is available (the suite gates on server availability and skips otherwise).

### DIFF
--- a/NArk.Swaps/Abstractions/SwapAsset.cs
+++ b/NArk.Swaps/Abstractions/SwapAsset.cs
@@ -8,4 +8,12 @@ public record SwapAsset(SwapNetwork Network, string AssetId)
 
     public static SwapAsset ArkAsset(string assetId)
         => new(SwapNetwork.Ark, assetId);
+
+    /// <summary>
+    /// ERC-20 token asset on an EVM chain — used by cross-chain providers
+    /// (LendaSwap) where <paramref name="contractAddress"/> is the 0x-prefixed
+    /// token contract address on the chosen <paramref name="chain"/>.
+    /// </summary>
+    public static SwapAsset Erc20(SwapNetwork chain, string contractAddress)
+        => new(chain, contractAddress);
 }
--- a/NArk.Swaps/Abstractions/SwapNetwork.cs
+++ b/NArk.Swaps/Abstractions/SwapNetwork.cs
@@ -5,4 +5,7 @@ public enum SwapNetwork
     Ark,
     BitcoinOnchain,
     Lightning,
+    EvmEthereum,
+    EvmPolygon,
+    EvmArbitrum,
 }
--- a/NArk.Swaps/Hosting/SwapServiceCollectionExtensions.cs
+++ b/NArk.Swaps/Hosting/SwapServiceCollectionExtensions.cs
@@ -6,6 +6,8 @@ using NArk.Swaps.Abstractions;
 using NArk.Swaps.Boltz;
 using NArk.Swaps.Boltz.Client;
 using NArk.Swaps.Boltz.Models;
+using NArk.Swaps.LendaSwap;
+using NArk.Swaps.LendaSwap.Client;
 using NArk.Swaps.Policies;
 using NArk.Swaps.Recovery;
 using NArk.Swaps.Services;
@@ -77,4 +79,20 @@ public static class SwapServiceCollectionExtensions
         return services;
     }
 
+    /// <summary>
+    /// Registers the LendaSwap provider and its dependencies.
+    /// Can be called separately if using manual provider registration.
+    /// </summary>
+    public static IServiceCollection AddLendaSwapProvider(
+        this IServiceCollection services, Action<LendaSwapOptions>? configure = null)
+    {
+        if (configure != null)
+            services.Configure(configure);
+
+        services.AddHttpClient<LendaSwapClient>();
+        services.AddSingleton<LendaSwapProvider>();
+        services.AddSingleton<ISwapProvider>(sp => sp.GetRequiredService<LendaSwapProvider>());
+
+        return services;
+    }
 }
--- a/NArk.Swaps/LendaSwap/Client/LendaSwapClient.Quotes.cs
+++ b/NArk.Swaps/LendaSwap/Client/LendaSwapClient.Quotes.cs
@@ -1,0 +1,52 @@
+using System.Net.Http.Json;
+using NArk.Swaps.LendaSwap.Models;
+
+namespace NArk.Swaps.LendaSwap.Client;
+
+public partial class LendaSwapClient
+{
+    /// <summary>
+    /// Lists available trading pairs and tokens.
+    /// </summary>
+    public virtual async Task<TokenListResponse?> GetTokensAsync(CancellationToken ct = default)
+    {
+        return await _httpClient.GetFromJsonAsync<TokenListResponse>("tokens", JsonOptions, ct);
+    }
+
+    /// <summary>
+    /// Gets a swap quote for the specified route and amount.
+    /// </summary>
+    /// <param name="sourceChain">Source chain identifier (e.g. "Arkade", "Lightning", "Bitcoin", "137").</param>
+    /// <param name="sourceToken">Source token identifier.</param>
+    /// <param name="targetChain">Target chain identifier.</param>
+    /// <param name="targetToken">Target token identifier.</param>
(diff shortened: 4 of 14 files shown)


ghost (changes requested):
## Arkana Code Review — LendaSwap Provider

Good structure overall. The partial-class client, route mapping, and test coverage are solid. But there are two real bugs and several issues that need fixing before merge.

---

### 🔴 BUG: `StopAsync` doesn't actually stop anything

**`LendaSwapProvider.cs:124-127`** — `StopAsync` logs and returns `Task.CompletedTask`. It never cancels `_shutdownCts`, so the polling loop continues running indefinitely after the host calls `StopAsync`.

Compare with `BoltzSwapProvider.StopAsync` which calls `ShutdownAsync()` → cancels the CTS → awaits background tasks. The `ISwapProvider` interface doc explicitly says *"cancel background tasks and await them before returning"*.

**Fix:** Mirror the Boltz pattern — extract a `ShutdownAsync()` method, call it from both `StopAsync` and `DisposeAsync`, make it idempotent with `Interlocked.Exchange`:

```csharp
public Task StopAsync(CancellationToken ct) => ShutdownAsync();

private int _shutdownStarted;
private async Task ShutdownAsync()
{
    if (Interlocked.Exchange(ref _shutdownStarted, 1) == 1) return;
    await _shutdownCts.CancelAsync();
    if (_pollingTask is not null)
        await _pollingTask;
}
```

---

### 🔴 BUG: `CancellationTokenSource` leak in `StartAsync`

**`LendaSwapProvider.cs:117-118`** — `CreateLinkedTokenSource` allocates a registration handle that must be disposed. `linkedCts` is never stored or disposed, leaking the handle for the lifetime of the process.

**Fix:** Store it as a field and dispose in `ShutdownAsync`/`DisposeAsync`, or just pass `_shutdownCts.Token` directly to the polling loop (the `ct` parameter from `StartAsync` is the host startup token — it signals *startup cancellation*, not *shutdown*. You should not link it to the polling loop's lifetime. Use `_shutdownCts.Token` alone.)

---

### 🟡 Null-forgiving `!` on deserialization

**`LendaSwapClient.cs:36`** — `ReadFromJsonAsync<TReturn>` can return `null` (empty body, `"null"` literal). The `!` suppresses the warning but gives you an `NRE` at the call site instead of a meaningful error.

```csharp
// Current:
return (await resp.Content.ReadFromJsonAsync<TReturn>(options: JsonOptions, ct))!;

// Better:
var result = await resp.Content.ReadFromJsonAsync<TReturn>(options: JsonOptions, ct);
return result ?? throw new HttpRequestException(
    $"LendaSwap returned empty response for {uri}", null, resp.StatusCode);
```

---

### 🟡 `Erc20()` factory accepts non-EVM networks

**`SwapAsset.cs:15`** — Nothing stops `SwapAsset.Erc20(SwapNetwork.Lightning, "0xabc")`. This will produce a semantically invalid asset that passes route matching in unexpected ways.

Add a guard:
```csharp
public static SwapAsset Erc20(SwapNetwork chain, string contractAddress)
{
    if (chain is not (SwapNetwork.EvmEthereum or SwapNetwork.EvmPolygon or SwapNetwork.EvmArbitrum))
        throw new ArgumentOutOfRangeException(nameof(chain), chain, "Erc20 requires an EVM network");
    return new(chain, contractAddress);
}
```

---

### 🟡 No backoff on polling failures

**`LendaSwapProvider.cs:180-230`** — If LendaSwap API is down, every 30s cycle logs an error per active swap with no backoff. With 100 active swaps and an hour-long outage, that's ~12,000 log lines.

Consider exponential backoff on consecutive cycle failures (cap at e.g. 5 min), reset on success.

---

### 🟢 Minor

- **`LendaSwapTests.cs:17-20`** (E2E): `CreateClient()` sets `HttpClient.BaseAddress` then the constructor overwrites it from `ApiUrl`. First assignment is dead code — remove it.
- **`LendaSwapTests.cs:48`** (E2E): `tokens.BtcTokens` dereferences a nullable `TokenListResponse?`. Will NPE if the API returns null. Add a null-check or use `Assert.That(tokens, Is.Not.Null)` before accessing members (already done on line 48, but `tokens!.BtcTokens` would be safer).
- **Status mapping question**: `clientredeemedandclientrefunded` → `Failed`. The name implies the client *did* redeem successfully. Is this really a failure from the user's perspective, or a partial success? Worth double-checking with the LendaSwap API docs.

---

### Summary

The architecture and test coverage are good. The two `StopAsync`/CTS bugs are the blockers — they cause resource leaks and violate the `ISwapProvider` lifecycle contract. Fix those, address the null-safety issue, and this is ready.

⚠️ Not protocol-critical: This PR doesn't touch VTXO handling, transaction signing, forfeit paths, or exit paths. The swap creation flows go through LendaSwap's hosted service (custodial bridge), not on-chain script execution in the SDK.

PR #39, opened 2026-03-17
feat: unilateral exit support

Kukks:
## Summary

- **Virtual Tx Storage**: Per-tx normalized storage model for VTXO exit data. Sibling VTXOs sharing internal tree nodes deduplicate automatically. Supports Lite (txids only) and Full (txids + hex) modes.
- **Unilateral Exit Service**: Full exit state machine — broadcasts chain of virtual txs root-to-leaf via 1p1c package relay, waits CSV timelock, then claims funds on-chain using the UnilateralPath tapscript.
- **Exit Watchtower**: Monitors blockchain for partial tree broadcasts and alerts when auto-exit is needed to protect funds from previous owners in the chain.
- **P2A CPFP Builder**: Builds v3 CPFP child transactions spending P2A anchor outputs (both BIP 431 standard and Ark protocol marker).
- **IOnchainBroadcaster**: Interface with NBXplorer (Bitcoin Core RPC) and Esplora (REST API) implementations for tx broadcast, package relay, status queries, and fee estimation.
- **Event Integration**: Auto-fetches exit data on VTXO receive, auto-prunes on VTXO spend.

## New files (36 changed, +2577 lines)

**Abstractions**: `IOnchainBroadcaster`, `IVirtualTxStorage`, `IExitSessionStorage`, `ExitSession`, `VirtualTx`, `VtxoBranch`  
**Core Services**: `VirtualTxService`, `UnilateralExitService`, `ExitWatchtowerService`, `P2ACpfpBuilder`  
**Transport**: `GetVtxoChainAsync`, `GetVirtualTxsAsync`, `GetVtxoTreeAsync` (gRPC + REST)  
**Storage**: `VirtualTxEntity`, `VtxoBranchEntity`, `ExitSessionEntity` (EF Core)  
**Tests**: 12 new unit tests for VirtualTxService and P2ACpfpBuilder

## Test plan

- [x] All 255 unit tests pass (`dotnet test NArk.Tests`)
- [x] `dotnet build NArk.Core` — 0 errors
- [x] `dotnet build NArk.Storage.EfCore` — 0 errors
- [ ] E2E tests with nigiri (future PR)
- [ ] Integration test with arkd v0.9 (future PR)


Kukks:
@arkanaai  what happens if we dont have enough onchain funds to complete unilateral exit? Is the unilateral exit system aware that only confirmed utxos can be used in a package? and that it has essential step and wait with every broadcast?

ghost (changes requested):
## 🔍 Arkana Review — New Commits `ae5354d0..9038a8da` (E2E tests)

Four commits adding E2E coverage for the unilateral exit pipeline. **Tests themselves are well-structured** — good state-machine assertions, proper timeout handling via `[CancelAfter]`, idempotency check, CSV delay maturation check. The iterative debugging across commits (`c7d9e99a → 6e20c8f4 → 9038a8da`) is transparent and the final state works.

**However, the test iteration uncovered a production bug that needs to be fixed in this PR or tracked explicitly.**

---

### 🔴 Production Bug: `PostBatchVtxoPollingHandler` misses VTXOs on inactive contracts

**File:** `NArk.Core/Events/PostBatchVtxoPollingHandler.cs:50`

```csharp
var contracts = await contractStorage.GetContracts(
    walletIds: [walletId],
    isActive: true,   // ← filters out inactive contracts
    cancellationToken: cancellationToken);
```

Commit `9038a8da`'s own message documents the problem clearly:

> *"SimpleIntentScheduler derives the SendToSelf output contract as Inactive [...] so the SDK's PostBatchVtxoPollingHandler — which filters isActive=true — was skipping the script the new VTXO actually lands on."*

The test works around this by replacing the handler entirely with an `InlineEventHandler` that polls ALL contracts. But the **production code still has this bug** — any user of the SDK using the standard DI registration (`AddArkServices`) will hit the same issue: after a batch, the new offchain VTXO won't be detected by the post-batch handler, and the user's balance won't update until the next routine poll cycle (or never, if the contract stays inactive).

This is not just a cosmetic issue — if the VTXO isn't detected, it can't be spent or exited, and could be silently lost to a round expiry.

**Fix:** Either:
- (a) Remove the `isActive: true` filter in `PostBatchVtxoPollingHandler` (poll all wallet contracts), or
- (b) Have `SimpleIntentScheduler` derive the SendToSelf contract as Active, or  
- (c) Pass both active and newly-derived contract scripts from the batch event itself

### 🟡 Tests bypass the production handler — production path untested

Because the inline handler replaces `PostBatchVtxoPollingHandler`, the actual production code path (`PostBatchVtxoPollingHandler.HandleAsync`) is never exercised in E2E. Once the above bug is fixed, the E2E should use the real handler to validate the fix.

### 🟡 Kukks' open questions not addressed in code

The maintainer raised two protocol-level concerns in the PR comments:

1. **0-value P2A direct broadcast rejection**: On non-regtest networks, `feeWallet: null` (the default) means every exit silently retries to death. This is documented in my earlier review and your response, but no code change was made (e.g., throwing at registration, logging a startup warning, or defaulting `IFeeWallet` to required).

2. **Confirmed UTXO enforcement**: `IFeeWallet.SelectFeeUtxoAsync` documents "confirmed" but doesn't enforce it. An implementor returning an unconfirmed UTXO will cause `submitpackage` to fail at the node.

These should either be addressed in code or explicitly tracked as follow-up issues with links.

### ✅ What's good

- **Test structure**: `SettleAVtxoAsync` helper is clean — boards, settles, wires exit deps, returns a disposable. Good pattern for other E2E suites to follow.
- **State machine coverage**: Broadcasting → AwaitingCsvDelay transition, CSV delay enforcement (no premature advance), idempotent start — these are the right tests to write.
- **Timeout discipline**: `[CancelAfter(180_000)]` / `[CancelAfter(240_000)]` with `CancellationToken` threading through all async calls. Proper.
- **Event-driven VTXO detection**: Using `VtxosChanged` event + TCS instead of polling is the right approach. The triple-TCS synchronization (`newSuccessBatch` → `batchPolledTcs` → `settledVtxoTcs`) correctly sequences the async pipeline.
- **Diagnostic output**: `TestContext.WriteLine` with state/retry/fail info on each step — will make CI failures debuggable.

### Minor

- **`UnilateralExitTests.cs:327`**: The `500ms` delay comment says "enough on regtest" — consider making this configurable or using the same `BatchSuccessPollingDelay` option that the production handler uses, so the test mirrors prod behavior.
- **`ExitTestSetup` doesn't dispose the `ServiceProvider`** — minor leak in test context, but `sp` (the built `ServiceProvider`) is never disposed. The `IAsyncDisposable[]` only covers the background services.

---

**Verdict:** Request changes — the `isActive: true` filter bug in `PostBatchVtxoPollingHandler` is a real production issue exposed by the test iteration. Fix it (or explicitly track it) before merge. The E2E tests are otherwise solid work.

*— Arkana 🔍*

ghost (approved):
## 🔍 Arkana Review — New Commits `9038a8da..0ee7084e`

Eight commits: PSBT parsing fix, whole-chain-with-types storage, opt-in auto-fetch refactor, commitment-row skip in broadcasting, E2E test hardening. **All previous findings addressed. Approving with minor notes below.**

---

### ✅ PSBT Parsing Fix — Correct (`5b916fec`, `5f1d1c0d`)

`ParseVirtualTx` now parses as PSBT and lifts `FinalScriptWitness`/`FinalScriptSig` directly — the right approach given arkd's encoding. The null-guard on each PSBT input field is correct defensive coding.

### ✅ Whole-Chain Storage with `ChainedTxType` — Clean (`2dad1633`)

Storing the full chain including the commitment anchor is the right call — consumers can walk back to the on-chain root without re-querying the indexer. The type enum in `NArk.Abstractions` (not `Transport.Models`) is correctly placed for cross-layer access.

### ✅ Commitment Skip in Broadcasting (`06cad189`)

Both `StartExitAsync` (hex-null check) and `ProgressBroadcastingAsync` (loop skip) correctly handle commitment rows. The `Unspecified` type falls through to the strict path (treated as needing hex) — fail-loud is the right default for a protocol-critical flow.

### ✅ `VtxoChainAutoFetchService` — Well-Designed (`2dad1633`)

The move from event-handler-with-retry-loop to `Channel<T>`-backed background service is a significant improvement. `SingleReader = true` is correct for the sequential processing pattern. `UnboundedChannel` is acceptable here — VTXO arrival rate is naturally bounded by batch frequency.

---

### 🟡 Minor: `EnsureHexPopulatedAsync` Omits `Type` in Update Records

**`VirtualTxService.cs` ~line 145:**
```csharp
updates.Add(new VirtualTx(txids[i], hexList[i], missingHex[i].ExpiresAt));
```

This defaults `Type` to `Unspecified`. The upsert logic in `EfCoreVirtualTxStorage` guards against overwriting with `Unspecified`, so this **won't clobber** existing types today. But it's fragile — if the upsert guard is ever changed, this silently resets types. Should be:
```csharp
updates.Add(new VirtualTx(txids[i], hexList[i], missingHex[i].ExpiresAt, missingHex[i].Type));
```

### 🟡 Missing EF Core Migration for `Type` Column

`VirtualTxEntity` gained a `Type` property with `HasConversion<int>()` and `HasDefaultValue(ChainedTxType.Unspecified)`. No migration was added. Consumers using relational DBs will hit a runtime column-not-found error until they generate one. At minimum, document this as a required migration step in the PR description, or ship the migration.

### 🟡 Broadcasting Blockers Are Tracked?

The `[Ignore]` annotations on two E2E tests document two distinct arkd-side issues:
1. Tree-tx PSBTs from `GetVirtualTxs` lack `FinalScriptWitness` → empty witnesses → mempool rejection
2. First tree-tx is v3 (TRUC) but parent is non-v3 → `TRUC-violation`

These block the entire broadcasting path on non-regtest. Are these tracked as arkd issues? Without a fix on the server side, unilateral exit is inoperable end-to-end.

---

### Summary

The new commits are solid. The PSBT parsing fix is a real correctness improvement, whole-chain storage is the right architecture, and the opt-in auto-fetch is cleaner than the old event handler. The two minor code items (Type propagation, missing migration) are low-risk. The broadcasting blockers are server-side and correctly deferred.

**Approve** — pending human sign-off on the protocol-critical exit path (this is an L2 unilateral exit pipeline).

*— Arkana 🔍*

ghost (changes requested):
## 🔍 Arkana Review — New Commit `0ee7084e..2494bd6f`
*— Arkana 🔍*

### 🚨 CRITICAL: Wrong `WitScript` constructor — malformed witness (protocol-critical)

**`NArk.Core/Services/UnilateralExitService.cs:656`**

```csharp
tx.Inputs[i].WitScript = new WitScript(sig.ToBytes());
```

This uses the `WitScript(byte[])` constructor which interprets the bytes as **serialized witness data** (varint count + length-prefixed stack items), NOT as a single witness stack push.

Every other taproot key-path spend in the codebase correctly uses:
```csharp
new WitScript(new[] { sig.ToBytes() }, true)
```

Evidence:
- `P2ACpfpBuilder.cs:94` — `new WitScript(new[] { sig.ToBytes() }, true)`
- `ChainSwapMusigSession.cs:104,212` — `new WitScript(new[] { finalSig.ToBytes() }, true)`
- `UnilateralExitService.cs:493-495` (claim tx, same file!) — `new WitScript([sig.ToBytes(), ...], true)`

**Fix:**
```csharp
tx.Inputs[i].WitScript = new WitScript(new[] { sig.ToBytes() }, true);
```

Without this fix, the tree-tx witness will be malformed and Bitcoin Core will reject the broadcast with a script verification failure. **This is the exit path for user funds** — if it doesn't work, users cannot unilaterally exit.

---

### ⚠️ Tests un-ignored but TRUC-violation still unresolved

The XML doc on `ProgressExits_AdvancesFromBroadcastingToAwaitingCsvDelay` (line 145-148) still documents a known TRUC-violation issue:

> The first tree-tx is v3 (TRUC) but its parent is non-v3, tripping `TRUC-violation`.

The `[Ignore]` attribute was removed, but this commit only fixes the witness assembly — it does NOT address the v3/TRUC relay policy issue. Either:
1. The TRUC issue was already fixed elsewhere (if so, remove the stale XML doc), or
2. These tests will still fail on Bitcoin Core 28+ nodes with TRUC enforcement, in which case keep `[Ignore]` until the tree-tx version or commitment-tx version is resolved upstream.

Which is it? The test remarks are now contradictory with the test being enabled.

---

### Summary

The witness fix approach is correct in spirit (lifting `TaprootKeySignature` for Tree txs matches the ts-sdk reference at `unroll.ts:183-189`), but the wrong constructor makes it produce garbage on the wire. One-line fix, then this is good.

**Reviewed by: Arkana (PR Reviewer Agent)**

ghost (changes requested):
## 🔍 Arkana Review — New Commit `0ee7084e..2494bd6f`
*— Arkana 🔍*

### 🚨 CRITICAL: Wrong `WitScript` constructor — malformed witness (protocol-critical)

**`NArk.Core/Services/UnilateralExitService.cs:656`**

```csharp
tx.Inputs[i].WitScript = new WitScript(sig.ToBytes());
```

This uses the `WitScript(byte[])` constructor which interprets the bytes as **serialized witness data** (varint count + length-prefixed stack items), NOT as a single witness stack push.

Every other taproot key-path spend in the codebase correctly uses:
```csharp
new WitScript(new[] { sig.ToBytes() }, true)
```

Evidence:
- `P2ACpfpBuilder.cs:94` — `new WitScript(new[] { sig.ToBytes() }, true)`
- `ChainSwapMusigSession.cs:104,212` — `new WitScript(new[] { finalSig.ToBytes() }, true)`
- `UnilateralExitService.cs:493-495` (claim tx, same file!) — `new WitScript([sig.ToBytes(), ...], true)`

**Fix:**
```csharp
tx.Inputs[i].WitScript = new WitScript(new[] { sig.ToBytes() }, true);
```

Without this fix, the tree-tx witness will be malformed and Bitcoin Core will reject the broadcast with a script verification failure. **This is the exit path for user funds** — if it doesn't work, users cannot unilaterally exit.

---

### ⚠️ Tests un-ignored but TRUC-violation still unresolved

The XML doc on `ProgressExits_AdvancesFromBroadcastingToAwaitingCsvDelay` (line 145-148) still documents a known TRUC-violation issue:

> The first tree-tx is v3 (TRUC) but its parent is non-v3, tripping `TRUC-violation`.

The `[Ignore]` attribute was removed, but this commit only fixes the witness assembly — it does NOT address the v3/TRUC relay policy issue. Either:
1. The TRUC issue was already fixed elsewhere (if so, remove the stale XML doc), or
2. These tests will still fail on Bitcoin Core 28+ nodes with TRUC enforcement, in which case keep `[Ignore]` until the tree-tx version or commitment-tx version is resolved upstream.

Which is it? The test remarks are now contradictory with the test being enabled.

---

### Summary

The witness fix approach is correct in spirit (lifting `TaprootKeySignature` for Tree txs matches the ts-sdk reference at `unroll.ts:183-189`), but the wrong constructor makes it produce garbage on the wire. One-line fix, then this is good.

**Reviewed by: Arkana (PR Reviewer Agent)**

ghost (approved):
## 🔍 Arkana Review — New Commits `b6dda81f..32cc19e6`
*— Arkana 🔍*

Reviewed 3 non-merge commits since last review:

### ✅ `eda81d2e` — fix(test): decode UnilateralExit Sequence properly in CSV-delay test

Correct fix. BIP68 time-based sequences (bit 22 set) produce raw `.Value` in the millions — mining that many regtest blocks was wrecking bitcoind/LND/Boltz for the rest of the suite. Now gates on `LockType`: height-based mines `LockHeight+2`, time-based early-returns after validating the don't-advance-early assertion. Pragmatic; the time-based CSV maturation via `setmocktime`+MTP is correctly deferred as separate work.

### ✅ `70a4f6cc` — fix(exit): make EF entity registration opt-in + default to Lite mode

Two good changes:

1. **Opt-in `ConfigureArkExitEntities()`** — follows the existing `ConfigureArkPaymentEntities` pattern exactly. Consumers who never call `AddUnilateralExit` no longer get three unwanted migration tables. Clean.

2. **Default `VirtualTxMode.Lite`** — correct default. Full mode does 2 arkd round-trips per VTXO arrival and most VTXOs never exit unilaterally. Lite defers hex fetch to `StartExitAsync`. Docs clearly explain the trade-off and the opt-in path for offline-exit requirements.

**Minor note (non-blocking):** All three `Configure*` methods instantiate independent `ArkStorageOptions`. If a consumer passes `configure: o => o.Schema = "foo"` to `ConfigureArkEntities` but forgets to pass it to `ConfigureArkExitEntities`, exit tables land in the default schema while core tables are in `"foo"`. This is a pre-existing pattern from `ConfigureArkPaymentEntities` (not a regression), but worth a doc note or a future refactor to a shared options builder.

### ✅ `32cc19e6` — fix(tests): TestDbContext opts in to exit entities

Trivially correct — mirrors the sample wallet's `WalletDbContext` change from the prior commit.

### Verdict

All three commits are clean. No protocol-critical concerns. No security issues. No cross-repo breakage (the `ConfigureArkExitEntities` is additive and opt-in). LGTM.

PR #77, opened 2026-04-25
feat: HD wallet gap-limit recovery via modular discovery providers

Kukks:
## Summary

Re-importing an HD wallet from its mnemonic used to leave the local SDK with no record of the contracts the previous instance derived. This PR adds a gap-limit scanner that rebuilds that state by sweeping derivation indices and asking each registered `IContractDiscoveryProvider` whether it ever saw activity at that index — OR semantics across all sources.

## Why modular providers

The user had three different sources of truth in mind:

- **arkd indexer** — has the wallet's payment contracts ever held a VTXO?
- **on-chain (NBXplorer/Esplora)** — has any historical UTXO landed at the boarding address?
- **boltz** — has the index's user pubkey participated in a submarine/reverse swap?

Each runs over a different protocol. Making this pluggable means a custom integration can drop in a `MyEscrowDiscoveryProvider` without touching the orchestrator. The interface is small (`Task<DiscoveryResult> DiscoverAsync(wallet, descriptor, index, ct)`) and provider results are unioned.

## Surfaces

- `NArk.Abstractions.Recovery` — `IContractDiscoveryProvider`, `DiscoveryResult`, `RecoveryOptions`, `RecoveryReport`
- `NArk.Core.Recovery` — `HdWalletRecoveryService` orchestrator, `IndexerVtxoDiscoveryProvider`, `BoardingUtxoDiscoveryProvider` (conditional on `IBoardingUtxoProvider` being registered)
- `NArk.Swaps.Recovery` — `BoltzSwapDiscoveryProvider` (delegates to existing `SwapsManagementService.RestoreSwaps` so VHTLC + swap rows get the canonical `Source=swap:<id>` metadata)
- DI: registered automatically by `AddArkCoreServices` + `AddArkSwapServices`

## Behaviour

- Discovered contracts are persisted with `Source=recovery:<provider>` + `RecoveryIndex=<n>` metadata.
- Dedup by script — multiple providers reconstructing the same contract → only one persisted.
- `wallet.LastUsedIndex` is bumped to `HighestUsedIndex + 1` but never lowered.
- Idempotent — safe to re-run.
- `SingleKey` wallets throw (no derivation index).

## Tests

12 new test cases in `NArk.Tests/Recovery/HdWalletRecoveryServiceTests.cs` cover:
- No usage / usage at 0 / interleaved usage
- OR semantics across two providers
- Throwing provider treated as not-found
- Per-script dedupe
- `StartIndex` / `MaxIndex` honoured
- `SingleKey` and unknown-wallet rejection
- Null-provider filtering
- Never-lower-`LastUsedIndex` invariant

Also fixes two pre-existing stale tests in `VtxoPollingHandlerTests` that were left broken on master by the v2.1.5 `GetVtxoByScriptsAsSnapshot` signature change and the `b014d1f` one-by-one polling removal.

`dotnet test NArk.Tests` → 269/269 pass.

## Docs + sample wallet

- README "HD Wallet Recovery" section
- New `docs/articles/recovery.md` with usage, custom-provider how-to, and tuning notes — registered in `toc.yml`
- Sample wallet (`samples/NArk.Wallet`) gets a "Restore from mnemonic / nsec" panel on the home page that runs the scanner and shows the discovered-contract count

DocFX build green (1 pre-existing unrelated warning on `index.md`).

## Test plan
- [ ] `dotnet test NArk.Tests` is green (269/269)
- [ ] `dotnet build NArk.sln` is green including the sample wallet
- [ ] DocFX build is green
- [ ] Smoke-test the sample wallet: paste a known-active mnemonic into the restore panel and confirm contracts get listed and balances appear

### DIFF
--- a/NArk.Abstractions/Recovery/IContractDiscoveryProvider.cs
+++ b/NArk.Abstractions/Recovery/IContractDiscoveryProvider.cs
@@ -1,0 +1,71 @@
+using NArk.Abstractions.Contracts;
+using NArk.Abstractions.Wallets;
+using NBitcoin.Scripting;
+
+namespace NArk.Abstractions.Recovery;
+
+/// <summary>
+/// Probes a single derivation index of an HD wallet to determine whether any
+/// contract derivable from that index has ever been used. Each implementation
+/// answers from a different source of truth (arkd indexer, on-chain UTXO set,
+/// boltz swap history, etc.) and they are aggregated by
+/// <c>HdWalletRecoveryService</c> using a logical OR — if any provider sees
+/// usage at an index, the index counts as used and the gap counter resets.
+/// </summary>
+/// <remarks>
+/// Providers are stateless and may be queried concurrently for different
+/// indices. They MUST NOT mutate <see cref="IWalletStorage"/> or
+/// <see cref="IContractStorage"/> directly; instead, return any contracts they
+/// reconstructed via <see cref="DiscoveryResult.Contracts"/>, and the
+/// orchestrator persists them once recovery completes.
+/// </remarks>
+public interface IContractDiscoveryProvider
+{
+    /// <summary>
+    /// Short identifier used in log lines and the recovery report
+    /// (e.g. <c>"indexer"</c>, <c>"boarding"</c>, <c>"boltz"</c>).
+    /// </summary>
+    string Name { get; }
+
+    /// <summary>
+    /// Probe whether any contract derivable from <paramref name="userDescriptor"/>
+    /// at <paramref name="index"/> has ever been used.
+    /// </summary>
+    /// <param name="wallet">The HD wallet being recovered.</param>
+    /// <param name="userDescriptor">
+    /// The output descriptor at the given index — concrete (no remaining
+    /// <c>/*</c>), already derived from <c>wallet.AccountDescriptor</c>.
+    /// </param>
+    /// <param name="index">The derivation index being probed.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>
+    /// A <see cref="DiscoveryResult"/> describing whether the index was used and,
+    /// if so, the reconstructed contracts to persist.
+    /// </returns>
+    Task<DiscoveryResult> DiscoverAsync(
+        ArkWalletInfo wallet,
+        OutputDescriptor userDescriptor,
+        int index,
+        CancellationToken cancellationToken = default);
+}
+
+/// <summary>
+/// Result of querying a single <see cref="IContractDiscoveryProvider"/> for one index.
+/// </summary>
+/// <param name="Used">
+/// <c>true</c> if the provider found evidence the descriptor at this index has
+/// been used (a VTXO, an on-chain boarding UTXO, a boltz swap, etc.).
+/// </param>
+/// <param name="Contracts">
+/// Contracts the provider reconstructed and would like the orchestrator to
+/// persist. May be empty even when <paramref name="Used"/> is true if the
+/// provider only knows "this was used" but not enough to materialize a
+/// contract on its own.
+/// </param>
+public record DiscoveryResult(
+    bool Used,
+    IReadOnlyList<ArkContract> Contracts)
+{
+    /// <summary>Convenience: provider saw nothing at this index.</summary>
+    public static DiscoveryResult NotFound { get; } = new(false, []);
+}
--- a/NArk.Abstractions/Recovery/IContractDiscoveryProvider.cs
+++ b/NArk.Abstractions/Recovery/IContractDiscoveryProvider.cs
@@ -13,11 +13,29 @@ namespace NArk.Abstractions.Recovery;
 /// usage at an index, the index counts as used and the gap counter resets.
 /// </summary>
 /// <remarks>
+/// <para>
 /// Providers are stateless and may be queried concurrently for different
-/// indices. They MUST NOT mutate <see cref="IWalletStorage"/> or
-/// <see cref="IContractStorage"/> directly; instead, return any contracts they
-/// reconstructed via <see cref="DiscoveryResult.Contracts"/>, and the
-/// orchestrator persists them once recovery completes.
+/// indices (and concurrently with each other for the same index).
+/// </para>
+/// <para>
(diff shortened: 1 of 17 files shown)


ghost (changes requested):
## Code Review — arkade-os/dotnet-sdk#77

Solid design. The modular provider pattern is clean, the interface is minimal, tests are thorough (12 cases covering the important edge cases), and the OR-semantics + dedup-by-script + never-lower-LastUsedIndex invariants are all correct. Documentation is good.

That said, I have findings across three severity tiers. None are show-stoppers but two need fixing before merge.

---

### 🔴 Must fix

**1. `DeriveAtIndex` diverges from canonical derivation logic**
`HdWalletRecoveryService.cs:161-168`

```csharp
var resolved = accountDescriptor.Contains("/0/*")
    ? accountDescriptor.Replace("/0/*", $"/0/{index}")
    : accountDescriptor.Replace("/*", $"/{index}");
```

The canonical implementation in `HierarchicalDeterministicAddressProvider.GetDescriptorFromIndex` (line 58-61) is simply:
```csharp
return OutputDescriptor.Parse(descriptor.Replace("/*", $"/{index}"), network);
```

The recovery code adds a special case for `/0/*` that doesn't exist in the production derivation path. If `AccountDescriptor` ever contains `/0/*`, the two methods will derive **different scripts for the same index**, and recovery will look at wrong addresses — silently missing funds.

Either: (a) call the existing `GetDescriptorFromIndex` (make it `internal static` and reuse), or (b) drop the `/0/*` branch and match the canonical logic exactly. Divergent derivation logic in a Bitcoin wallet is a correctness risk.

**2. `GetServerInfoAsync` called per-index in both providers — O(n) unnecessary network round-trips**
`IndexerVtxoDiscoveryProvider.cs:30` and `BoardingUtxoDiscoveryProvider.cs:22`

Both providers call `await clientTransport.GetServerInfoAsync()` inside `DiscoverAsync`, which runs once per derivation index. For a gap-20 scan with usage at index 500, that's ~520 × 2 = 1040 extra gRPC calls that all return the same `ArkServerInfo`.

The orchestrator (`HdWalletRecoveryService.ScanAsync`) already fetches `serverInfo` at line 51. Pass it into `DiscoverAsync` as a parameter (extend the interface), or at minimum cache it inside each provider instance. As-is, a full 10k-index scan would make **20,000 redundant network calls**.

---

### 🟡 Should fix

**3. No input validation on `RecoveryOptions`**
`RecoveryOptions.cs:22-24`

`GapLimit`, `MaxIndex`, and `StartIndex` are all unconstrained `int`. Negative `StartIndex` or `GapLimit ≤ 0` would produce nonsensical behavior (infinite loop or immediate exit). Add guard clauses:
```csharp
public record RecoveryOptions(int GapLimit = 20, int MaxIndex = 10_000, int StartIndex = 0)
{
    public RecoveryOptions
    {
        ArgumentOutOfRangeException.ThrowIfNegativeOrZero(GapLimit);
        ArgumentOutOfRangeException.ThrowIfNegative(StartIndex);
        ArgumentOutOfRangeException.ThrowIfNegative(MaxIndex);
        if (StartIndex > MaxIndex) throw new ArgumentOutOfRangeException(nameof(StartIndex));
    }
}
```

**4. Sequential provider queries — easy parallelism left on the table**
`HdWalletRecoveryService.cs:82-100`

Providers are queried with a sequential `foreach` per index. Since providers hit different backends (gRPC to arkd, HTTP to Boltz, HTTP to Esplora), parallelizing them with `Task.WhenAll` per index would cut latency by ~3x for a typical 3-provider setup. The providers are documented as stateless and safe for concurrent use.

**5. `BoltzSwapDiscoveryProvider` mutates storage inside `DiscoverAsync` — violates interface contract**
`BoltzSwapDiscoveryProvider.cs:25-26` and `IContractDiscoveryProvider.cs:19-21`

The interface doc says: *"They MUST NOT mutate IWalletStorage or IContractStorage directly."* But `RestoreSwaps` persists VHTLC contracts and swap rows inside the call. The PR description and the provider's own doc comment explain *why* this is intentional (canonical swap metadata), but it contradicts the interface contract. Either update the interface doc to note this exception, or restructure so `RestoreSwaps` is called post-scan by the orchestrator (which would be cleaner — the orchestrator already has a persist phase).

---

### 🟢 Nits / observations

**6. `serverKey` param in `PersistDiscoveriesAsync` passed to `ToEntity` but `ToEntity` ignores it**
`HdWalletRecoveryService.cs:174` → `ArkContract.ToEntity` accepts `defaultServerKey` but the method body never uses it in the `ArkContractEntity` constructor. This is a pre-existing issue in the SDK, not introduced by this PR, but it means the `serverKey` plumbing is dead code.

**7. Recovered contracts marked `ContractActivityState.Active`**
`HdWalletRecoveryService.cs:192`

If a recovered contract was already fully spent, it'll be marked Active until VTXO polling reconciles. This is fine in practice (polling will correct it), but worth a comment explaining the intent.

**8. Sample wallet `_restoreSecret` — mnemonic in Blazor WASM memory**
`Home.razor:161` — Mnemonic lives in a C# string in browser memory. Standard WASM risk; just note it in the sample's README since this is a reference implementation people will copy.

**9. Test fix for `VtxoPollingHandlerTests` is correct**
The 4-param `GetVtxoByScriptsAsSnapshot` overload update and the `Received(1)` assertion change (batch vs one-by-one) match the v2.1.5 signature change. Good cleanup.

---

### Protocol-critical flag 🚨

This PR touches **contract derivation and persistence** — the `DeriveAtIndex` method determines which scripts are watched for VTXOs, and `PersistDiscoveriesAsync` writes contract state that drives balance calculations. While the recovery path doesn't sign transactions, **incorrect derivation silently loses visibility of funds**. Finding #1 (divergent derivation logic) must be resolved before merge.

**Requesting human review** per protocol-critical code policy.

ghost (approved):
## Re-review — arkade-os/dotnet-sdk#77 (commit `812c45c`)

Only change since last approval (`e6df6ea`): submodule pointer `regtest` pinned to `cca4fba` (CI-only, no source changes).

Previous approval stands. Still protocol-critical (HD derivation + contract persistence) — **human review still required**.

PR #78, opened 2026-04-25
feat: per-wallet VTXO sync cursor in wallet metadata

Kukks:
## Summary

After a process restart, `VtxoSynchronizationService` had no idea what it had already synced and re-fetched the full historical VTXO set for every active script. For wallets with thousands of historical VTXOs that's a multi-second hit on every container restart, plus a wave of arkd indexer load. This PR adds a persistent cursor stored as a per-wallet metadata entry so cold-start catch-up only fetches changes since the last shutdown.

## Storage shape — generic JSONB on `ArkWalletEntity`

Earlier iterations of this PR used a dedicated `ArkSyncStateEntity` singleton table — replaced after pushback on the heaviness. The cursor now rides on a new generic `Metadata: Dictionary<string,string>?` column on `ArkWalletEntity` (provider-agnostic JSON via EF value-converter — Postgres `jsonb`, SQLite `TEXT`, SQL Server `nvarchar(max)` all work).

`IWalletStorage.SetMetadataValue(walletId, key, value, ct)` does sparse-key updates so future per-wallet bookkeeping (recovery state, etc.) can use the same column without colliding with the sync cursor.

## Mechanism

| Poll source | Subset | After-filter on cold-start | Advances cursor on success |
|---|---|---|---|
| `UpdateScriptsView` initial catch-up (StartAsync) | All active scripts | **`MIN(per-wallet vtxo.lastFullPollAt)`** (or `null` if any wallet has no cursor — fresh-wallet safety) | no |
| `UpdateScriptsView` follow-up (new contract added at runtime) | Just the newly-added scripts | `null` (genuinely new) | no |
| `RoutinePoll` 5-second tick | All active scripts (full-set snapshot) | `UtcNow - 2 min` | **yes** — writes `StartedAt` to every wallet |
| Stream push retries (750ms / 3s / 8s) | Scripts arkd flagged | `UtcNow - 5 min` | no |

Cursor written = poll's `StartedAt` (not finish time) — guarantees any VTXO change that lands during the in-flight poll is still inside the next `after` window.

### Gap-loss safety net

A `_coldStartCatchupComplete` flag gates the routine-poll cursor writes until the cold-start catch-up has succeeded at least once. Without it: a transient catch-up failure followed by a routine-poll success would advance the cursor past the catch-up window and permanently skip any VTXO that landed during the downtime.

### Fresh-wallet safety

The cold-start `MIN` returns `null` if any wallet has no `vtxo.lastFullPollAt` entry — a fresh wallet's first-time scripts must trigger full-history catch-up rather than skipping their window via someone else's cursor.

## Surface

- `ArkWalletInfo.Metadata: IReadOnlyDictionary<string,string>?` (new, optional, defaults to null)
- `ArkWalletEntity.Metadata: Dictionary<string,string>?` (new JSON column with `ValueComparer` so EF tracks dictionary mutations correctly)
- `IWalletStorage.SetMetadataValue(walletId, key, value, ct)` (new, sparse update; passing `value=null` removes the key)
- `VtxoSynchronizationService.LastFullPollAtMetadataKey = "vtxo.lastFullPollAt"` (constant, exported alongside the key-owning service)
- `VtxoSynchronizationService` constructor gains an optional `IWalletStorage` parameter; if absent, falls back to current behaviour (null-after on cold start)

## No downstream migration burden

The previous version of this PR required consumer apps (BTCPay plugin) to add an `AddSyncState` migration on SDK bump. **That requirement is gone.** Adding a column to an existing `Wallets` table is still a migration on the consumer side, but it's a 1-column add against the existing entity rather than a new singleton table — and there's no new DI wiring at all.

## Removed since the previous review pass

- `NArk.Abstractions/Sync/ISyncStateStorage.cs`
- `NArk.Storage.EfCore/Entities/ArkSyncStateEntity.cs`
- `NArk.Storage.EfCore/Storage/EfCoreSyncStateStorage.cs`
- `ArkStorageOptions.SyncStateTable`
- 2 DI lines in `AddArkEfCoreStorage`
- `ArkSyncStateEntity.Configure()` call in `ConfigureArkEntities`
- Migration callout in `docs/articles/storage.md` (replaced with one paragraph on the wallet-metadata cursor approach)
- `EfCoreSyncStateStorageTests.cs`

## Tests

- `EfCoreWalletStorageTests` (10 total, 6 new):
  - Adds key to empty metadata
  - Preserves unrelated keys across multiple writers (concurrency / no-clobber check)
  - Overwrites existing key
  - `value=null` removes the key
  - Removing the last key sets `Metadata` back to `null`
  - Throws on unknown wallet
- `VtxoSynchronizationServiceSyncStateTests` (rewritten for `IWalletStorage` shape, 6 cases):
  - Cold-start uses MIN-cursor across wallets as `after`
  - Cold-start with any missing cursor → null `after` (fresh-wallet safety)
  - Cold-start with zero wallets → null `after`
  - Opt-out: `walletStorage=null` still works
  - Successful catch-up writes the cursor to every wallet
  - Failed catch-up never writes the cursor (gap-loss safety net)

`dotnet test NArk.Tests` → 269/269 green locally.

## Test plan

- [x] `dotnet test NArk.Tests` is green
- [x] `dotnet build NArk.sln` is clean
- [ ] CI green
- [ ] BTCPay plugin migration adds the new `Metadata` column on bump


ghost (changes requested):
## Review: feat: persistent VTXO sync-state cursor bounds cold-start catch-up

**Reviewer: Arkana (automated aggressive review)**
**Verdict: Request changes — protocol-critical code touching VTXO sync requires human sign-off**

---

### Overall Assessment

This is a well-structured, low-risk quality-of-life improvement. The mechanism is sound: persist the `LastFullPollAt` cursor so cold restarts don't refetch the entire VTXO history. The new `ISyncStateStorage` abstraction is clean, the optional constructor parameter preserves backward compatibility, and the test coverage is reasonable. I have one actionable concern and a few minor notes, but the main reason for requesting changes is protocol policy: this touches VTXO synchronization, which is the mechanism by which wallets discover their funds — **a human must sign off**.

---

### ✅ What looks good

1. **Cursor semantics are correct.** Advancing to `StartedAt` (not completion time) guarantees no gap between poll windows. The 2-minute `RoutinePollLookback` provides generous overlap.

2. **Backward compatibility is clean.** `ISyncStateStorage? syncStateStorage = null` in both constructors means all existing callers (7 direct `new VtxoSynchronizationService(...)` call sites in tests, DI via `AddArkCoreServices`, btcpay-arkade injection) continue working without changes.

3. **DI wiring is correct.** `AddArkEfCoreStorage` registers `EfCoreSyncStateStorage → ISyncStateStorage`, and since `VtxoSynchronizationService` is resolved via `services.AddSingleton<VtxoSynchronizationService>()` in `AddArkCoreServices`, the DI container will inject the optional `ISyncStateStorage` when both registration sets are used together.

4. **Singleton + DbContextFactory pattern is correct.** `EfCoreSyncStateStorage` is registered as Singleton but creates a fresh `DbContext` per operation via `IArkDbContextFactory` — no scoped-lifetime mismatch.

5. **Channel serialization eliminates cursor write races.** `StartQueryLogic` processes `PollRequest`s sequentially from a single `Channel<PollRequest>` reader, so `SetLastFullPollAtAsync` is never called concurrently. No lock needed.

6. **`_isFirstStartupCatchup` is protected by `_viewSyncLock`.** Both `StartAsync → UpdateScriptsView` and `OnActiveScriptsChanged → UpdateScriptsView` acquire the semaphore, so the flag can't be consumed by a race.

7. **Fail-safe on bad `StartedAt`.** Non-`IsFullSetSnapshot` requests default `StartedAt` to `default(DateTimeOffset)` (0001-01-01). If a bug accidentally flips `IsFullSetSnapshot=true`, the cursor regresses to epoch, making the next cold-start do a full-history fetch. That's safe.

---

### ⚠️ Actionable: gap-loss on transient cold-start catch-up failure

**`VtxoSynchronizationService.cs` — interaction between `UpdateScriptsView` (line ~290) and `StartQueryLogic` (line ~403)**

The cold-start catch-up enqueues a `PollRequest` with `IsFullSetSnapshot=false`. If that poll **fails** in `StartQueryLogic` (transient indexer/network error), the error is logged and the loop continues. Five seconds later, the first `RoutinePoll` succeeds and writes `IsFullSetSnapshot=true` — advancing the cursor to `~UtcNow`.

The gap between `stored-cursor` and `UtcNow - 2min` is now permanently skipped. VTXOs that landed in that window are never fetched, and no subsequent mechanism will retry them. The old behavior (no cursor) would retry the full-history fetch on next restart.

**Severity: Low-medium.** Requires: (a) downtime > 2 minutes, AND (b) the catch-up poll fails, AND (c) the immediately-following routine poll succeeds. Unlikely in practice, but this is a Bitcoin L2 — "unlikely" ≠ acceptable when missing a VTXO means invisible funds.

**Suggested mitigation — pick one:**
- **(A) Don't advance the cursor until the catch-up succeeds.** Add a `_coldStartCatchupComplete` flag; skip cursor advancement in `StartQueryLogic` until it's true. Set it when the first cold-start catch-up request completes without error.
- **(B) Make the cold-start catch-up `IsFullSetSnapshot=true`.** If it fails, the cursor isn't advanced (failure path skips the `SetLastFullPollAtAsync` call). If it succeeds, the cursor advances to `startedAt` which is correct — the catch-up covered everything from `stored-cursor` onward.
- **(C) Retry the catch-up before allowing routine polls.** Simplest but delays startup.

I'd lean toward **(A)** — minimal diff and explicit.

---

### 📝 Minor notes (non-blocking)

1. **`EfCoreSyncStateStorage.SetLastFullPollAtAsync` — no explicit transaction.** The read-then-write (lines 29-40) isn't wrapped in a transaction. Since writes are serialized by the Channel, this is safe in practice, but a `db.Database.BeginTransactionAsync` would make the upsert robust against any future caller that bypasses the channel.

2. **Missing test: cursor advancement after successful full-set poll.** The 3 new `VtxoSynchronizationServiceSyncStateTests` verify cold-start *reads* but none verify that `SetLastFullPollAtAsync` is called after a successful `IsFullSetSnapshot=true` poll. Consider adding a test that triggers a RoutinePoll-like request and asserts `_syncState.Received().SetLastFullPollAtAsync(...)`.

3. **Docs: migration instructions are clear.** The `dotnet ef migrations add AddSyncState` callout is good. The btcpay-arkade team should be aware — they call `AddArkEfCoreStorage<ArkPluginDbContext>` in `ArkPlugin.cs:121` and will need a migration on the next SDK bump.

4. **Cross-repo impact: none detected.** `ISyncStateStorage` has zero consumers outside this repo currently. No other SDKs (ts-sdk, go-sdk, rust-sdk) are affected. No breaking API changes.

---

### 🔒 Protocol-critical flag

This PR modifies VTXO synchronization — the mechanism by which wallets discover their funds. While the change is well-designed and the risk is bounded, **a human reviewer must approve before merge per Arkana protocol rules**. The gap-loss scenario described above, while unlikely, could cause a wallet to not see funds that arrived during a restart window.

ghost (approved):
## Follow-up review: `de98193b` — gate cursor advance on cold-start catch-up success

**Reviewer: Arkana (automated aggressive review)**
**Verdict: Approve — the gap-loss concern from the previous review is resolved**

---

### What changed

This commit implements option (A) from my previous review: a `_coldStartCatchupComplete` gate that prevents routine polls from advancing the persisted cursor until the cold-start catch-up has succeeded at least once.

### Verification

1. **Gate logic is correct.** `_coldStartCatchupComplete` starts `false` (or `true` when `syncStateStorage is null` — preserving backward compat). The catch-up poll carries `IsColdStartCatchup=true`. On success, the gate flips *before* the cursor-advance check. Routine polls (`IsFullSetSnapshot=true, IsColdStartCatchup=false`) are blocked until that flip. ✅

2. **Catch-up now also advances the cursor immediately.** The catch-up `PollRequest` is `IsFullSetSnapshot=true` with `StartedAt=DateTimeOffset.UtcNow`, so on its successful completion the cursor advances right there — no 5-second window waiting for the first routine poll. This is cleaner than the original design. ✅

3. **`volatile` on `_coldStartCatchupComplete` is sufficient.** The flag is written once (catch-up success in `StartQueryLogic`) and read on subsequent iterations of the same sequential Channel reader loop. `volatile` ensures visibility. No lock needed. ✅

4. **Failure scenario is fully gated.** If catch-up fails → `_coldStartCatchupComplete` stays `false` → routine poll succeeds → `IsFullSetSnapshot && false && ...` → cursor NOT advanced. The gap is preserved for the next restart. ✅

5. **Test coverage is solid.** `ColdStart_OnSuccessfulCatchup_AdvancesCursor` verifies the happy path. `ColdStart_OnCatchupFailure_DoesNotAdvanceCursor` directly tests the gap-loss scenario. The `Throwing()` async-iterator helper is the standard C# pattern for this. ✅

6. **`VtxoPollingHandlerTests` signature fix.** The `GetVtxoByScriptsAsSnapshot` mock calls now include the `DateTimeOffset?` parameters added in the first commit. These were presumably broken before this commit — good catch. ✅

7. **Docs updated.** `storage.md` now correctly describes the gated advancement behavior. ✅

### No new concerns

All items from the previous review are addressed. The implementation is clean and minimal.

---

### 🔒 Protocol-critical flag (unchanged)

This PR still modifies VTXO synchronization — the mechanism by which wallets discover their funds. **A human reviewer must approve before merge.**

ghost (approved):
## Follow-up review: `e88b4729`..`dbdada92` — CI-only changes

**Reviewer: Arkana (automated aggressive review)**
**Verdict: Approve maintained**

---

Two commits since my last review (`de98193b`):

| Commit | Description |
|--------|-------------|
| `e88b4729` | CI re-trigger after stuck runner |
| `dbdada92` | Pin `regtest` submodule to `cca4fba` (known-working with PR #79's E2E) |

**Diff: submodule pointer bump only.** No application code, no test code, no workflow changes. Zero impact on the sync-state cursor logic reviewed previously.

Previous approval stands. 🔒 Protocol-critical flag unchanged — human sign-off still required before merge.

PR #84, opened 2026-05-05
feat(logging): tag swap + batch log entries with WalletId scope

Kukks:
## Summary

Adds `using (logger.BeginScope(("WalletId", id)))` at every per-wallet entry point in the SDK so downstream sinks (e.g. the BTCPay plugin's per-wallet diagnostic log added in [ArkLabsHQ/btcpay-arkade#46](https://github.com/ArkLabsHQ/btcpay-arkade/pull/46)) can route every transitively-emitted log line to the right wallet — including ones whose message template doesn't already carry `{WalletId}`.

## Why

Without scopes, capture is best-effort: a custom `ILoggerProvider` can only see `WalletId` if the call site happens to include it as a structured arg. Many services log heavily but with `{SwapId}` / `{IntentId}` / `{Outpoint}` only — so per-wallet capture missed them.

Adding a scope is cheap (one allocation per iteration / call) and propagates through async via `AsyncLocal`, so any sub-method called from the scoped region inherits it without further changes.

## What's scoped

**`NArk.Swaps/Services/SwapsManagementService.cs`**
- `PollSwapState` foreach iteration body (after the swap is loaded) — transitively covers `TryClaimBtcForChainSwap`, `TrySignBoltzBtcClaim`, `RequestRefundCooperatively`
- Public initiate / pay / restore methods: `InitiateSubmarineSwap`, `PayExistingSubmarineSwap`, `InitiateReverseSwap`, `InitiateBtcToArkChainSwap`, `InitiateArkToBtcChainSwap`, `RestoreSwaps`

**`NArk.Core/Services/BatchManagementService.cs`**
- `RouteToBatchSessionsAsync` per-intent loop — transitively covers `HandleBatchFailedAsync`, `HandleBatchFinalizedAsync`, `HandleBatchExceptionAsync`
- `HandleBatchStartedForAllIntentsAsync` per-intent loop — transitively covers `SetupBatchSessionAsync`

**`NArk.Core/Services/OnchainService.cs`**
- `InitiateCollaborativeExit(string walletId, …)` and the `(ArkCoin[] inputs, …)` overload (the latter derives the wallet from `inputs[0].WalletIdentifier` and opens the scope before validations so rejection logs carry it too)

**`NArk.Core/Services/IntentGenerationService.cs`**
- `GenerateManualIntent(string walletId, …)`
- The periodic per-wallet loop in the generation cycle (`foreach (var walletContracts in contracts)`) — `using var` at the top of each iteration disposes between wallets

**`NArk.Core/Services/SpendingService.cs`**
- `Spend(string walletId, ArkCoin[], …)`, `Spend(string walletId, ArkTxOut[], …)`, and `GetAvailableCoins(string walletId, …)`

**`NArk.Core/Services/AssetManager.cs`**
- `IssueAsync(string walletId, …)`, `ReissueAsync(string walletId, …)`, `BurnAsync(string walletId, …)`

**`NArk.Core/Recovery/HdWalletRecoveryService.cs`**
- `ScanAsync(string walletId, …)`

**`NArk.Core/Services/DelegationMonitorService.cs`**
- `ProcessVtxoAsync(ArkVtxo vtxo)` — scope opened right after `walletId` is resolved from `contract.WalletIdentifier`

**`NArk.Core/Services/SweeperService.cs`**
- `Sweep(HashSet<ArkCoin>)` per-coin loop — the coin set may span multiple wallets, so the scope is per iteration, not per call

## What still isn't scoped (and why)

`VtxoSynchronizationService` polls / streams span **all** active wallets at once. A single log line can pertain to many wallets, so a per-wallet scope would be misleading. Sinks that want per-wallet routing of indexer activity will need a different mechanism (e.g. inspecting the script→wallet map at log time).

## Behaviour

No behavioural changes. The `using (...)` blocks only add scope context to the existing async flow. `continue` inside the new `using` blocks still targets the surrounding `foreach` (the using's finally disposes the scope before the iteration exits).

## Test plan

- [x] `dotnet test NArk.Tests` — 273 / 273 pass
- [ ] CI: build + unit + e2e green

### DIFF
--- a/NArk.Core/Recovery/HdWalletRecoveryService.cs
+++ b/NArk.Core/Recovery/HdWalletRecoveryService.cs
@@ -45,6 +45,7 @@ public class HdWalletRecoveryService(
         RecoveryOptions? options = null,
         CancellationToken cancellationToken = default)
     {
+        using var _walletScope = logger?.BeginScope(("WalletId", walletId));
         options ??= RecoveryOptions.Default;
         options.Validate();
 
--- a/NArk.Core/Services/AssetManager.cs
+++ b/NArk.Core/Services/AssetManager.cs
@@ -34,6 +34,7 @@ public class AssetManager(
     public async Task<IssuanceResult> IssueAsync(string walletId, IssuanceParams parameters,
         CancellationToken cancellationToken = default)
     {
+        using var _walletScope = logger?.BeginScope(("WalletId", walletId));
         logger?.LogDebug("Issuing {Amount} new asset units for wallet {WalletId}", parameters.Amount, walletId);
 
         var serverInfo = await transport.GetServerInfoAsync(cancellationToken);
@@ -120,6 +121,7 @@ public class AssetManager(
     public async Task<string> ReissueAsync(string walletId, ReissuanceParams parameters,
         CancellationToken cancellationToken = default)
     {
+        using var _walletScope = logger?.BeginScope(("WalletId", walletId));
         logger?.LogDebug("Reissuing {Amount} units using control asset {AssetId} for wallet {WalletId}",
             parameters.Amount, parameters.AssetId, walletId);
 
@@ -195,6 +197,7 @@ public class AssetManager(
     public async Task<string> BurnAsync(string walletId, BurnParams parameters,
         CancellationToken cancellationToken = default)
     {
+        using var _walletScope = logger?.BeginScope(("WalletId", walletId));
         logger?.LogDebug("Burning {Amount} units of asset {AssetId} for wallet {WalletId}",
             parameters.Amount, parameters.AssetId, walletId);
 
--- a/NArk.Core/Services/BatchManagementService.cs
+++ b/NArk.Core/Services/BatchManagementService.cs
@@ -219,34 +219,40 @@ public class BatchManagementService(
             if (!_activeIntents.TryGetValue(intentId, out var intent))
                 continue;
 
-            try
+            // Tag every log emitted during this intent's batch event handling
+            // with the owning wallet so per-wallet diagnostic-log capture
+            // (in BTCPay plugin) can route them.
+            using (logger?.BeginScope(("WalletId", intent.WalletId)))
             {
-                var isComplete = await session.ProcessEventAsync(eventResponse, cancellationToken);
-                if (isComplete)
+                try
                 {
+                    var isComplete = await session.ProcessEventAsync(eventResponse, cancellationToken);
+                    if (isComplete)
+                    {
+                        CleanupBatchSession(intentId, batchId);
+                    }
+
+                    switch (eventResponse)
+                    {
+                        case BatchFailedEvent batchFailed when batchFailed.Id == intent.BatchId:
+                            await HandleBatchFailedAsync(intent, batchFailed, cancellationToken);
+                            CleanupBatchSession(intentId, batchId);
+                            _activeIntents.TryRemove(intentId, out _);
+                            _ = UpdateTopicsAsync(removeTopics: GetTopicsForIntent(intent));
+                            break;
+
+                        case BatchFinalizedEvent batchFinalized when batchFinalized.Id == intent.BatchId:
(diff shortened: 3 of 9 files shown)


ghost (approved):
## ✅ Approved — clean logging instrumentation

Reviewed the full diff across all 9 changed files. This is a pure observability improvement — no protocol logic, no behavioral changes, no public API surface modifications.

### What I verified

1. **`using var` with null-conditional is safe** — `using (IDisposable? x)` where `x` is null is a no-op in C#. All `logger?.BeginScope(...)` sites handle nullable logger correctly.

2. **`continue` inside `using` blocks** — confirmed the `using`'s `Dispose` fires before the `foreach` iteration exits. No scope leaks. This applies to both `SwapsManagementService.PollSwapState` and `BatchManagementService.RouteToBatchSessionsAsync`.

3. **`SwapsManagementService.PollSwapState` restructure** — the `using` scope is correctly placed *inside* the outer `try` block, after the swap is loaded from storage (when `WalletId` is known). The `BoltzSwapNotFoundException` and generic `Exception` catches remain *outside* the wallet scope, which is correct since the swap may not have been loaded yet at that point. `_scriptToSwapId` assignment moved inside the `using` block is pure code motion, no semantic change.

4. **`OnchainService.cs` multi-input overload** (`InitiateCollaborativeExit(ArkCoin[], ...)`) — `inputs.Length > 0` guard before accessing `inputs[0].WalletIdentifier`, and the null-conditional `using` handles the empty-inputs case. Safe.

5. **`SweeperService.cs`** — per-iteration scope is correct since the coin set may span multiple wallets. Good comment documenting this.

6. **No cross-repo impact** — no public API, type, or interface changes. Pure internal instrumentation. No consumers affected.

7. **No protocol-critical code touched** — no changes to VTXO handling, signing, forfeit paths, round lifecycle, or exit logic. The `OnchainService` collaborative exit changes are logging-only; the exit logic itself is untouched.

### Nits (non-blocking)

- This introduces `BeginScope` as a new pattern in the codebase (no prior usage). Might be worth a brief mention in contributing docs so future contributors know to add scopes at wallet-level entry points.

LGTM. Ship it.

ghost (approved):
## ✅ Approved — clean logging instrumentation

Reviewed the full diff across all 9 changed files. This is a pure observability improvement — no protocol logic, no behavioral changes, no public API surface modifications.

### What I verified

1. **`using var` with null-conditional is safe** — `using (IDisposable? x)` where `x` is null is a no-op in C#. All `logger?.BeginScope(...)` sites handle nullable logger correctly.

2. **`continue` inside `using` blocks** — confirmed the `using`'s `Dispose` fires before the `foreach` iteration exits. No scope leaks. This applies to both `SwapsManagementService.PollSwapState` and `BatchManagementService.RouteToBatchSessionsAsync`.

3. **`SwapsManagementService.PollSwapState` restructure** — the `using` scope is correctly placed *inside* the outer `try` block, after the swap is loaded from storage (when `WalletId` is known). The `BoltzSwapNotFoundException` and generic `Exception` catches remain *outside* the wallet scope, which is correct since the swap may not have been loaded yet at that point. `_scriptToSwapId` assignment moved inside the `using` block is pure code motion, no semantic change.

4. **`OnchainService.cs` multi-input overload** (`InitiateCollaborativeExit(ArkCoin[], ...)`) — `inputs.Length > 0` guard before accessing `inputs[0].WalletIdentifier`, and the null-conditional `using` handles the empty-inputs case. Safe.

5. **`SweeperService.cs`** — per-iteration scope is correct since the coin set may span multiple wallets. Good comment documenting this.

6. **No cross-repo impact** — no public API, type, or interface changes. Pure internal instrumentation. No consumers affected.

7. **No protocol-critical code touched** — no changes to VTXO handling, signing, forfeit paths, round lifecycle, or exit logic. The `OnchainService` collaborative exit changes are logging-only; the exit logic itself is untouched.

### Nits (non-blocking)

- This introduces `BeginScope` as a new pattern in the codebase (no prior usage). Might be worth a brief mention in contributing docs so future contributors know to add scopes at wallet-level entry points.

LGTM. Ship it.

PR #86, opened 2026-05-06
feat(swaps): plumb optional referralId on Boltz swap creation

Kukks:
## Summary

Mirrors [arkade-os/wallet#606](https://github.com/arkade-os/wallet/pull/606) (which added a \`referralId\` on the JS \`boltz-swap\` provider) on the .NET SDK side. Boltz uses the referralId to credit the originating integration for the swap; consumer apps pass their own identifier (the wallet uses \`arkade-money\`, the BTCPay plugin will use \`btcpay-arkade\` in a paired bump PR).

## Changes

- \`BoltzClientOptions\` gains an optional \`ReferralId\` property.
- \`BoltzClient\` exposes it as a public read-only property (so the swap orchestrator can stamp it without taking a second DI dependency).
- \`SubmarineRequest\` gains a \`ReferralId\` field. The other two swap-creation DTOs (\`ReverseRequest\`, \`ChainRequest\`) already had one — only Submarine was missing, so the JSON shape is now consistent across all three.
- \`BoltzSwapsService\` stamps \`boltzClient.ReferralId\` on every swap-create request: Submarine, Reverse, BTC→ARK chain, ARK→BTC chain.

## Surface change

Single field, opt-in, default null. Existing callers who don't configure a referral see no behavior change — \`JsonIgnoreCondition.WhenWritingNull\` (the existing client-level setting) keeps the field off the wire.

## Test plan

- [x] \`dotnet build NArk.sln\` — 0 errors
- [x] \`dotnet test NArk.Tests\` — 273/273 pass
- [ ] CI: build + unit + e2e green
- [ ] Downstream BTCPay PR sets \`ReferralId = "btcpay-arkade"\` and verifies a Submarine / Reverse / Chain swap goes out with that field

### DIFF
--- a/NArk.Swaps/Boltz/BoltzSwapsService.cs
+++ b/NArk.Swaps/Boltz/BoltzSwapsService.cs
@@ -40,6 +40,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
                 (extractedSender.PubKey?.ToBytes() ?? extractedSender.XOnlyPubKey.ToBytes()).ToHexStringLower(),
             From = "ARK",
             To = "BTC",
+            ReferralId = boltzClient.ReferralId,
         }, cancellationToken);
 
         if (invoice.PaymentHash is null)
@@ -99,6 +100,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
             DescriptionHash = createInvoiceRequest.DescriptionHash?.ToString(),
             Description = createInvoiceRequest.Description,
             InvoiceExpirySeconds = Convert.ToInt32(createInvoiceRequest.Expiry.TotalSeconds),
+            ReferralId = boltzClient.ReferralId,
         };
 
         var response = await boltzClient.CreateReverseSwapAsync(request, cancellationToken);
@@ -185,7 +187,8 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
             PreimageHash = Encoders.Hex.EncodeData(preimageHash),
             ClaimPublicKey = claimPubKeyHex,
             RefundPublicKey = Encoders.Hex.EncodeData(ephemeralKey.PubKey.ToBytes()),
-            ServerLockAmount = amountSats
+            ServerLockAmount = amountSats,
+            ReferralId = boltzClient.ReferralId,
         };
 
         var response = await boltzClient.CreateChainSwapAsync(request, ct);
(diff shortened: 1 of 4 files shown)


ghost (approved):
## ✅ Approved — clean, low-risk change

Reviewed: full diff (4 files, +26/−3), all four full source files, cross-repo consumers (`btcpay-arkade`), sibling request DTOs (`ReverseRequest`, `ChainRequest`), and the `BoltzClient` serialization pipeline.

### Verdict

This is a metadata/attribution field for Boltz swap-fee crediting. **Not protocol-critical** — no VTXO, signing, forfeit, round, or exit code is touched. The VHTLC contract construction, address validation, and preimage handling are all unchanged.

### What I verified

1. **All four swap-creation paths stamped** — `CreateSubmarineSwap`, `CreateReverseSwap`, `CreateBtcToArkSwapAsync`, `CreateArkToBtcSwapAsync` all set `ReferralId = boltzClient.ReferralId`. No path missed.

2. **Null-safety on the wire** — `BoltzClient` uses `JsonSerializerOptions` with `DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull` globally, so a null `ReferralId` is omitted from the JSON payload. Existing callers who don't configure a referral see zero wire change.

3. **Consistency with sibling DTOs** — `ReverseRequest` already had `ReferralId` (no explicit `[JsonIgnore]`), `ChainRequest` already had it (with explicit `[JsonIgnore(WhenWritingNull)]`). `SubmarineRequest` now matches `ReverseRequest`'s style. The global serializer setting covers both.

4. **Backward compatibility** — `BoltzClientOptions.ReferralId` is optional (`string?`), no `required` modifier. Existing DI registrations that don't set it get `null` → field omitted from wire → no behavior change.

5. **Cross-repo impact** — `btcpay-arkade` consumes `BoltzClient` via DI but doesn't reference `BoltzClientOptions` directly (it's configured via the NArk service registration). The paired BTCPay PR mentioned in the description will set the `ReferralId` on the options. No breakage.

### Nits (non-blocking)

- **`SubmarineRequest.cs:19`** — File still ends without a trailing newline (`\ No newline at end of file`). Trivial, but worth fixing while you're here.

- **No unit test for referral plumbing** — Consider a test that configures `BoltzClientOptions.ReferralId = "test"`, creates a swap request via the service, and asserts the request DTO carries the value. Not blocking since this is a pure pass-through, but it'd catch regressions if the property chain is refactored.

LGTM. Ship it.

🤖 Review by Arkana

PR #90, opened 2026-05-10
feat(recovery): reconcile pending Arkade txs stranded between Submit and Finalize

Kukks:
## Summary

Arkade off-chain transactions follow a two-phase **Submit → Finalize** flow. If the process crashes between phases, the server holds the inputs as in-flight and only allows the original pending tx to be finalized — without recovery, those coins are stuck indefinitely.

This PR adds `PendingArkTransactionRecoveryService`, which on host startup (via `ArkHostedLifecycle`, after `VtxoSync`), and on demand per-wallet:

1. Builds a BIP-322 ownership proof anchored on each batch's first VTXO (same shape as `GetIntentsByProofAsync`, with a `{"type":"get-pending-tx","expire_at":0}` envelope — matches `go-sdk` and `ts-sdk`).
2. Calls a new `IClientTransport.GetPendingTxAsync` endpoint to retrieve every pending tx the server is still holding for that identity. gRPC + REST + caching transports all wired up.
3. Resolves each checkpoint PSBT input back to a local `ArkCoin` via `IVtxoStorage` + `ICoinService`, signs the spending witness with `IArkadeWalletSigner`, and calls `FinalizeTx`.
4. Per-tx failures are logged + raised on `RecoveryFailed` (an `EventHandler<PendingTxRecoveryFailureEventArgs>`) so subscribers can show a banner / ship telemetry — the loop never blocks, and the next host start retries leftovers.

Coins are batched in groups of 20 (server hard limit), arkTxIds are deduped across batches, and the proof set deliberately includes spent VTXOs (`includeSpent: true`) because the very VTXO we want to recover is reported as `spent_by_pending` by the time recovery runs.

### Bug found and fixed along the way

`IntentProofHelper.SignBip322Proof` reassigned its local `psbt` parameter to a new packet via `PSBT.FromTransaction(...).UpdateFrom(psbt)`, signed that, and discarded it on return — so callers held an unsigned, leaf-script-less PSBT. This was masked for `GetIntentsByProof` (arkd's intent verification is more permissive) but tripped over `GetPendingTx`, which strictly requires `TaprootLeafScript` on each input. The helper now returns the signed PSBT and the three callers (`CreateIntentOwnershipProofAsync`, `CreateGetPendingTxOwnershipProofAsync`, `DelegationMonitorService`) use the returned instance.

## What's wired

- `AddArkCoreServices` registers the service as a singleton.
- `ArkHostedLifecycle.StartAsync` calls `RecoverAllWalletsAsync` after `VtxoSync.StartAsync`.
- `FinalizePendingArkTransactionsAsync(walletId, ct)` is the explicit per-wallet entry point.

## Tests

**Unit (NArk.Tests, NSubstitute mocks):**
- `NoSpendableVtxos_SkipsTransport`
- `NoSignerForWallet_SkipsTransport`
- `NoResolvableCoins_SkipsTransport`
- `HappyPath_FinalizesAndReturnsArkTxId`
- `DedupesAcrossBatches` (21 coins → 2 proof batches → same txid → 1 finalize)
- `FinalizeFailure_RaisesEventAndContinues`
- `RecoverAllWallets_ContinuesAcrossWallets_OnFailure`

**E2E (NArk.Tests.End2End, real arkd):**
- `RecoversPendingTx_WhenFinalizeNeverFollowedSubmit` — a `CrashAfterSubmitTransport` decorator simulates a process crash by throwing on `FinalizeTx` after `SubmitTx` reaches arkd. Recovery against the real transport then retrieves the pending tx, finalizes it, and a second sweep is asserted to be a no-op (idempotency). The test polls for arkd's async `spent_by_pending` projection to catch up — production never races this because recovery runs at host startup, well after any crash.

## Docs

- README.md — new "Pending Arkade Transaction Recovery" section with a minimal example + `RecoveryFailed` snippet.
- docs/articles/pending-tx-recovery.md — full conceptual article with sequence diagram + setup + edge cases (incl. the same-process projection-race note).
- XML docs on every new public surface; DocFX build clean.
- Sample wallet (`samples/NArk.Wallet/.../ArkWalletService.cs`) injects the service and exposes `FinalizePendingArkTransactions(walletId)`.

## Test plan

- [x] `dotnet test NArk.Tests` — all 292 unit tests pass
- [x] `docfx docfx.json` — clean build (cref + linting clean for new APIs)
- [x] E2E in CI green (~7m, real arkd via nigiri)
- [x] Build & Unit Tests in CI green


ghost (changes requested):
## 🔍 Arkana Code Review — PR #90: Pending Ark Transaction Recovery

**Verdict: Request changes** — one startup-blocking bug, plus this is protocol-critical code (transaction signing, VTXO finalization) that requires human sign-off before merge.

Overall: solid design, well-tested, API shape matches go-sdk/ts-sdk exactly. The recovery flow is sound. But there's a gap in the "best-effort" promise that can take down the host.

---

### 🔴 BUG — `RecoverAllWalletsAsync` can block host startup

**`PendingArkTransactionRecoveryService.cs:144-167`**

The outer `try/catch` only catches `OperationCanceledException`:

```csharp
try
{
    var wallets = await walletStorage.LoadAllWallets(cancellationToken);
    foreach (var wallet in wallets) { /* inner try/catch is fine */ }
}
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
{
    // host shutting down
}
```

If `walletStorage.LoadAllWallets()` throws anything else (database timeout, connection refused, corrupted state), it propagates uncaught through `ArkHostedLifecycle.StartAsync` and **kills the host on startup**. The comment on `ArkHostedLifecycle.cs:22` says "Best-effort: failures inside here are absorbed and logged so they never block the host from starting" — but the code doesn't deliver on that promise.

**Fix:** Add a general `catch (Exception ex) when (ex is not OperationCanceledException)` around the entire body of `RecoverAllWalletsAsync`, matching the pattern already used inside the per-wallet loop.

---

### 🟡 PROTOCOL — `checkpoint.Inputs.Single()` hard assumption

**`PendingArkTransactionRecoveryService.cs:285`**

```csharp
var inputPrevOut = checkpoint.Inputs.Single().PrevOut;
```

This assumes each checkpoint PSBT has exactly one input. If the server ever returns a checkpoint with zero or multiple inputs, `Single()` throws `InvalidOperationException` which is caught by the per-tx handler — so it won't crash — but it will silently fail to recover that tx with an unhelpful error message.

Consider: either add a guard with a clear log message ("checkpoint has {n} inputs, expected 1"), or if this is a hard protocol invariant, add a comment citing the spec/proto that guarantees it.

---

### 🟡 DEFENSIVE — REST transport empty-string fallback

**`RestClientTransport.Intents.cs:529-530`**

```csharp
var arkTxId = p.GetProperty("arkTxid").GetString() ?? string.Empty;
var finalArkTx = p.GetProperty("finalArkTx").GetString() ?? string.Empty;
```

If the server returns a null `arkTxid` for two different pending txs, both map to `""`, and the `HashSet<string> seen` dedup in the recovery service (line 253) would skip the second one. This is an unlikely server bug, but the consequence is silently dropping a pending tx from recovery.

Consider: skip entries where `arkTxId` is null/empty instead of coercing to `""`.

---

### 🟢 OBSERVATIONS (non-blocking)

1. **VTXO spent-marking timing is safe.** I verified that the SDK does NOT mark VTXOs as spent locally during SubmitTx/FinalizeTx — spent status flows from the server via polling (`PostSpendVtxoPollingHandler`). So after a crash between Submit and Finalize, local VTXOs remain "spendable" and recovery has proof material. ✅

2. **API shape matches cross-SDK.** `PendingArkTransaction(ArkTxId, FinalArkTx, SignedCheckpointTxs)` matches the proto `PendingTx` message in go-sdk and the `PendingTx` interface in ts-sdk. The BIP-322 message envelope `{"type":"get-pending-tx","expire_at":0}` matches both. ✅

3. **No cross-repo breakage.** `IClientTransport` is internal to dotnet-sdk; no other repos in the monorepo depend on it. The new `GetPendingTxAsync` method is additive. ✅

4. **Caching transport correctly passes through.** `GetPendingTxAsync` is not cached — correct, this must always hit the server for fresh state. ✅

5. **Test coverage is strong.** 7 unit tests cover: no VTXOs, no signer, no resolvable coins, happy path, cross-batch dedup, finalize failure + event, cross-wallet failure isolation. E2E test with `CrashAfterSubmitTransport` is a nice design. ✅

6. **`CreateStubServerInfo` in tests** (`PendingArkTransactionRecoveryServiceTests.cs:1077-1082`) uses `RuntimeHelpers.GetUninitializedObject` — functional but fragile. If `ArkServerInfo` ever adds constructor validation, this silently produces invalid state. Consider adding a test-friendly factory or using a builder pattern. Minor.

7. **Missing trailing newline** on `IClientTransport.cs` and `GrpcClientTransport.Txs.cs` (visible in diff as `\ No newline at end of file`). Trivial.

---

### ⚠️ PROTOCOL-CRITICAL FLAG

This PR touches **transaction signing** (`SignCheckpointAsync` → `PsbtHelpers.SignAndFillPsbt`), **VTXO finalization** (`FinalizeTx`), and **BIP-322 ownership proofs**. Per protocol-critical review policy: **requires explicit human sign-off before merge**, even after the bug above is fixed.

---

*🤖 Review by Arkana · arkade-os/dotnet-sdk#90*

ghost (approved):
## 🔍 Arkana Code Review — PR #90 (follow-up on `25984cc`)

All three issues from my previous review are fixed cleanly:

### ✅ BUG FIX — `RecoverAllWalletsAsync` no longer blocks host startup
`PendingArkTransactionRecoveryService.cs:97-106` — general `catch (Exception ex)` now wraps the entire method body, matching the "best-effort" contract. Log message is clear and actionable. New unit test (`RecoverAllWallets_AbsorbsWalletStorageFailure_SoHostStartupNeverBlocks`) pins the behavior. ✅

### ✅ PROTOCOL — `checkpoint.Inputs` guard
`PendingArkTransactionRecoveryService.cs:230-237` — explicit `Count != 1` check with an actionable error message citing the arkTxId and expected vs actual count. Switched from `.Single()` to `[0]` after the guard. Correct. ✅

### ✅ DEFENSIVE — REST transport null arkTxId
`RestClientTransport.Intents.cs:123-126` — null/empty `arkTxId` entries are now skipped with `continue` instead of coerced to `""`. Prevents silent dedup collisions in the recovery service's `HashSet<string>`. ✅

No new issues found in this commit. Code is ready from my side.

---

### ⚠️ PROTOCOL-CRITICAL — Human review still required

This PR touches transaction signing, VTXO finalization, and BIP-322 ownership proofs. Per protocol-critical review policy: **a human must approve before merge**.

*🤖 Review by Arkana · arkade-os/dotnet-sdk#90*

ghost (approved):
## 🔍 Arkana Code Review — PR #90 (follow-up on `25984cc`)

All three issues from my previous review are fixed cleanly:

### ✅ BUG FIX — `RecoverAllWalletsAsync` no longer blocks host startup
`PendingArkTransactionRecoveryService.cs:97-106` — general `catch (Exception ex)` now wraps the entire method body, matching the "best-effort" contract. Log message is clear and actionable. New unit test (`RecoverAllWallets_AbsorbsWalletStorageFailure_SoHostStartupNeverBlocks`) pins the behavior. ✅

### ✅ PROTOCOL — `checkpoint.Inputs` guard
`PendingArkTransactionRecoveryService.cs:230-237` — explicit `Count != 1` check with an actionable error message citing the arkTxId and expected vs actual count. Switched from `.Single()` to `[0]` after the guard. Correct. ✅

### ✅ DEFENSIVE — REST transport null arkTxId
`RestClientTransport.Intents.cs:123-126` — null/empty `arkTxId` entries are now skipped with `continue` instead of coerced to `""`. Prevents silent dedup collisions in the recovery service's `HashSet<string>`. ✅

No new issues found in this commit. Code is ready from my side.

---

### ⚠️ PROTOCOL-CRITICAL — Human review still required

This PR touches transaction signing, VTXO finalization, and BIP-322 ownership proofs. Per protocol-critical review policy: **a human must approve before merge**.

*🤖 Review by Arkana · arkade-os/dotnet-sdk#90*

PR #89, opened 2026-05-08
test(swaps): port unhappy-path E2E coverage from fulmine + boltz-swap

Kukks:
## Summary

Started as a port of unhappy-path swap E2E coverage from [ArkLabsHQ/fulmine](https://github.com/ArkLabsHQ/fulmine/tree/master/internal/test/e2e) and [arkade-os/boltz-swap](https://github.com/arkade-os/boltz-swap/tree/master/test/e2e). While porting, several real SDK feature gaps surfaced — chain-swap renegotiation, chain-swap cooperative refund (both directions), and recovery-inspection diagnostics — so the PR grew to close those alongside the test coverage.

## SDK feature additions

| Area | What landed | Why |
|---|---|---|
| **Chain-swap renegotiation** | `BoltzSwapProvider.PollSwapState` now handles `transaction.lockupFailed` by calling `BoltzClient.GetChainQuoteAsync` + `AcceptChainQuoteAsync` (these wire calls existed but had no consumer). On success the swap proceeds with the renegotiated `ExpectedAmount`; on Boltz refusal it falls through to refund. Mirrors `arkade-os/boltz-swap`'s `quoteSwap` and fulmine's `TestChainSwapBTCtoARKWithQuote`. |  Without this, every chain-swap funded with a non-exact amount went terminal Failed with funds stranded at the lockup. |
| **Chain-swap cooperative refund (BTC→ARK)** | New `CoopRefundBtcToArkChainSwap`: locates the BTC lockup tx via Boltz status, builds an unsigned refund tx, signs cooperatively via the existing `ChainSwapMusigSession.CooperativeRefundAsync`, broadcasts via Boltz's BTC broadcaster. | The signing primitive existed; nothing called it. Funds at a BTC lockup of a failed chain swap had no automated recovery. |
| **Chain-swap cooperative refund (ARK→BTC)** | Symmetric `CoopRefundArkToBtcChainSwap`: parses the VHTLC contract, snapshots arkd VTXOs at the contract script, builds an Ark refund tx, asks Boltz to co-sign via `RefundChainSwapArkAsync`, submits via the existing Ark transaction builder. | Same gap, opposite direction. |
| **Expired-with-no-funds → Failed safeguard** | When `swap.expired` arrives with no observable lockup, transition the swap to `Failed` instead of looping forever waiting to refund nothing. | Without this, the refund retry loop would keep an unfunded expired swap `Pending` indefinitely. |
[...]

PR #75, opened 2026-04-09
feat: payment tracking improvements

Kukks:
## Summary
- Add asset tracking to `ArkPayment` and `ArkPaymentRequest` (JSONB columns for asset lists)
- Add `Cancelled` status to both payment and payment request enums
- Convert `PaymentTrackingService` from constructor-subscription to proper `IHostedService`
- Add `SemaphoreSlim` serialization in `OnVtxoChanged` to prevent race conditions on `ReceivedAmount`
- Add `ArkBip21` fluent builder + parser (modeled after BTCPay's `ArkadeBip21Builder`)
- Add `MergeAssets` helper with duplicate-safe dictionary merging
- **Make payment tracking fully opt-in**: consumers call `AddArkPaymentTracking()` and `ConfigureArkPaymentEntities()` separately
- 35 new unit tests (27 BIP21 + 8 MergeAssets)

## Opt-in Usage
```csharp
// DI registration
services.AddArkEfCoreStorage<MyDbContext>();
services.AddArkPaymentTracking(); // opt-in

// DbContext
modelBuilder.ConfigureArkEntities();
modelBuilder.ConfigureArkPaymentEntities(); // opt-in
```

## Test plan
- [x] All 292 unit tests pass
- [x] Build succeeds across all projects (including WASM wallet sample)
- [ ] CI green (build + E2E)

PR #91, opened 2026-05-11
sample(wallet): dogfood ArkBip21.Parse + surface pending-tx + StillPending count

Kukks:
## Summary

Three drift fixes for the Blazor sample wallet to keep it current with what landed in PRs #75, #90, and #89.

1. **`Send.razor` uses `ArkBip21.Parse`** — replaces a 60+ line hand-rolled BIP21 / address parser with the SDK helper PR #75 shipped. LNURL still goes through `LnurlHelper` (the SDK deliberately doesn't resolve LNURL endpoints). Closes the dogfooding loop: the sample now exercises its own SDK's payment-string parser instead of competing with it.

2. **`Settings.razor` "Force pending-tx recovery" button** — surfaces the `FinalizePendingArkTransactions` API that PR #90 added on `ArkWalletService`. `ArkHostedLifecycle` already runs the sweep on startup, but a manual button is useful right after a crash if a tx looks stuck and you don't want to wait for the next restart.

3. **`Swaps.razor` shows `SwapRecoveryStatus.StillPending` count** — distinguishes "you have no active swaps" from "you have active swaps but recovery is correctly skipping them." Tells the user why a scan came up empty.

## Files

- `samples/NArk.Wallet/NArk.Wallet.Client/Pages/Send.razor` — drop custom `ParseBip21` (~60 lines), call `ArkBip21.Parse` + new `MapBip21PaymentInfo` helper. LNURL handling unchanged.
- `samples/NArk.Wallet/NArk.Wallet.Client/Pages/Settings.razor` — new Maintenance section with the recovery button + result message; calls `ArkWalletService.FinalizePendingArkTransactions`.
- `samples/NArk.Wallet/NArk.Wallet.Client/Pages/Swaps.razor` — show `StillPending` count alongside `Recoverable`.

## Test plan

- [x] \`dotnet build samples/NArk.Wallet/NArk.Wallet.Client/NArk.Wallet.Client.csproj\` — clean
- [x] \`dotnet test NArk.Tests\` — 349/349 pass
- [ ] CI build + E2E green

### DIFF
--- a/samples/NArk.Wallet/NArk.Wallet.Client/Pages/Send.razor
+++ b/samples/NArk.Wallet/NArk.Wallet.Client/Pages/Send.razor
@@ -5,6 +5,7 @@
 @inject IJSRuntime JS
 @inject LnurlHelper Lnurl
 @using NArk.Wallet.Client.Services
+@using NArk.Abstractions.Payments
 @using NArk.Swaps.Boltz
 
 <div class="page-header">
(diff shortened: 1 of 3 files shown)


ghost (changes requested):
## Code Review — arkade-os/dotnet-sdk#91

Good refactoring overall. Replacing the hand-rolled parser with `ArkBip21.Parse` is the right call. Two issues — one is a real bug, the other is dead code.

---

### 🐛 BUG: `StillPending` count is always zero (Swaps.razor)

**`Swaps.razor:26`** — `_recoveryReport.Count(r => r.Status == SwapRecoveryStatus.StillPending)` will always evaluate to `0`.

`ScanRecoverableSwapsAsync` (SwapsManagementService.cs:560-562) explicitly filters out pending swaps *before* calling `InspectSwapRecoveryAsync`:

```csharp
var candidates = allSwaps
    .Where(s => s.Status is not ArkSwapStatus.Pending)
    .ToList();
```

`InspectSwapRecoveryAsync` is the only place that produces `SwapRecoveryStatus.StillPending`, but it's never called for those swaps. The entire "still in flight" UI added in this PR is unreachable.

**Fix options (pick one):**
1. Count pending swaps directly from `_swaps` instead of `_recoveryReport`:
   ```csharp
   var stillPending = _swaps.Count(s => s.Status == ArkSwapStatus.Pending);
   ```
2. Or change `ScanRecoverableSwapsAsync` to include pending swaps in its output (would need a separate PR since that's SDK-level).

Option 1 is simpler and doesn't change the SDK contract.

---

### ⚠️ Dead code: `CollaborativeExit` switch arm (Send.razor)

**`Send.razor:293`** — `ArkPaymentMethod.CollaborativeExit => "onchain"` is dead code. `Bip21PaymentInfo.PreferredMethod` (Bip21PaymentInfo.cs:61-64) never returns `CollaborativeExit`:

```csharp
public ArkPaymentMethod PreferredMethod =>
    ArkAddress is not null || AssetId is not null ? ArkPaymentMethod.ArkSend :
    Lightning is not null ? ArkPaymentMethod.SubmarineSwap :
    OnchainAddress is not null ? ArkPaymentMethod.ChainSwap :
    ArkPaymentMethod.ArkSend;
```

Not a bug — it's defensive coding for a future change to `PreferredMethod`. But worth a `// defensive: PreferredMethod doesn't return this today` comment so the next reader doesn't wonder.

---

### ✅ Everything else looks good

- **Send.razor BIP21 migration**: Correct mapping. `ArkBip21.Parse` handles all the prefixes the old code did (`lntbs` is covered by `lntb` prefix match). The SDK's length check on legacy addresses is actually *better* than the old code (which would match `"1"` as an address).
- **Settings.razor recovery button**: Proper `_recoveryInFlight` guard against double-clicks, correct try/catch/finally pattern, safe string truncation on tx IDs.
- **No protocol-critical concerns**: All three files are sample wallet UI. No VTXO handling, signing, forfeit paths, or round lifecycle changes.
- **No cross-repo breakage**: No public API changes — this PR only consumes APIs from #75 and #90 (both merged).

---

**Verdict:** Request changes for the `StillPending` bug — the feature described in the PR summary ("distinguishes 'you have no active swaps' from 'you have active swaps but recovery is correctly skipping them'") doesn't actually work.

PR #92, opened 2026-05-12
fix(efcore): opt-in DateTimeOffset → long ticks mapping so SQLite ORDER BY works

Kukks:
## Problem

EF Core's SQLite provider rejects \`ORDER BY\` on \`DateTimeOffset\` columns:

> SQLite does not support expressions of type 'DateTimeOffset' in ORDER BY clauses. Convert the values to a supported type, or use LINQ to Objects to order the results on the client side.

This breaks every paged storage query in the SDK (\`GetVtxos\`, \`GetContracts\`, \`GetIntents\`, \`GetPayments\`, \`GetPaymentRequests\`, \`GetSwaps\`) when run against SQLite. Non-SQLite providers (Postgres / MSSQL) sort \`DateTimeOffset\` natively and aren't affected.

Reproduction: bundle NArk with a SQLite-backed \`ArkDbContext\` and call \`IVtxoStorage.GetVtxos(...)\` after any wallet activity. I hit this integrating NArk into Wasabi Wallet's Arkade flow — VTXO listing was unusable.

## Approach

Add \`ArkStorageOptions.StoreDateTimeOffsetAsTicks\` (default off). When on, applies a \`ValueConverter<DateTimeOffset, long>\` to every \`DateTimeOffset\` (and nullable) property in the Ark model — storage becomes BIGINT (Postgres/MSSQL) or INTEGER (SQLite), both natively sortable. Same on-disk size as the native types, fully indexable.

\`\`\`csharp
modelBuilder.ConfigureArkEntities(o => o.StoreDateTimeOffsetAsTicks = true);
\`\`\`

## Why an opt-in instead of always-on

I originally tried \"always-on\" but it's a schema change for *every* existing consumer regardless of provider:

- SQLite consumers: existing TEXT data wouldn't deserialize as long.
- Postgres/MSSQL consumers: existing \`timestamptz\` / \`datetimeoffset\` columns wouldn't load as BIGINT.

Default-off keeps that path safe. Consumers that hit the SQLite bug opt in explicitly and accept the trade-offs documented on the option:

1. Schema changes (TEXT/timestamp → BIGINT).
2. Round-trip strips the original offset (read-back is always UTC, offset zero) — the Ark entities use these columns as instants (\"when did this happen\") rather than zoned moments, so for the SDK's own usage this is fine.

Migration paths for existing SQLite consumers when enabling the flag:

1. Drop the DB file and let \`EnsureCreated\` rebuild with INTEGER columns (fine for local caches).
2. One-off SQL to convert TEXT columns to INTEGER ticks before flipping the flag.

## Tests

Added \`NArk.Tests/EfCoreSqliteOrderByTests.cs\`:

- \`TicksMapping_OrderByDateTimeOffset_WorksOnSqlite\` — proves the opt-in fix: \`ORDER BY CreatedAt DESC\` against SQLite returns rows in chronological order.
- \`TicksMapping_RoundTrip_StripsOffsetToUtc\` — pins the documented trade-off so it can't regress silently.
- \`DefaultMapping_StillWorksForReadAndWrite\` — backwards-compat: with the flag off, \`DateTimeOffset\` columns still store and read back correctly. (Only \`ORDER BY\` is broken on SQLite — the bug the opt-in fixes.)

All 352 tests pass (349 existing + 3 new).

## Out of scope

A future improvement could auto-detect SQLite and apply the conversion only there. That would still be a schema change for existing SQLite consumers though, so leaving the choice to the consumer feels right at this stage.

### DIFF
--- a/NArk.Storage.EfCore/ArkStorageOptions.cs
+++ b/NArk.Storage.EfCore/ArkStorageOptions.cs
@@ -28,4 +28,30 @@ public class ArkStorageOptions
     /// </summary>
     public Func<IQueryable<ArkWalletContractEntity>, string, IQueryable<ArkWalletContractEntity>>?
         ContractSearchProvider { get; set; }
+
+    /// <summary>
+    /// When true, stores every <see cref="DateTimeOffset"/> column as <see cref="long"/>
+    /// UTC ticks (BIGINT on Postgres/MSSQL, INTEGER on SQLite). Needed for SQLite consumers
+    /// because EF Core's SQLite provider rejects <c>ORDER BY</c> on the default TEXT
+    /// representation of <see cref="DateTimeOffset"/> — every paged query in this SDK
+    /// (<c>GetVtxos</c>, <c>GetContracts</c>, <c>GetIntents</c>, …) breaks otherwise.
+    ///
+    /// <para>
+    /// <b>Off by default</b> to preserve native column types (<c>timestamptz</c> / <c>datetimeoffset</c>)
+    /// for existing Postgres/MSSQL consumers. Enabling this is a schema change: stored values
+    /// switch from TEXT/timestamp to BIGINT and the original offset is dropped (read-back is
+    /// always UTC, offset zero). On-disk size is unchanged.
+    /// </para>
+    ///
+    /// <para>
+    /// Migration path for existing SQLite consumers:
+    /// <list type="number">
+    /// <item>If you can drop the DB (e.g. local cache), delete the file and let <c>EnsureCreated</c>
+    /// re-create the schema with INTEGER columns.</item>
+    /// <item>Otherwise run a one-off SQL migration to convert TEXT columns to INTEGER ticks
+    /// before enabling this flag.</item>
+    /// </list>
+    /// </para>
+    /// </summary>
+    public bool StoreDateTimeOffsetAsTicks { get; set; }
 }
--- a/NArk.Storage.EfCore/ModelBuilderExtensions.cs
+++ b/NArk.Storage.EfCore/ModelBuilderExtensions.cs
@@ -78,12 +78,35 @@ public static class ModelBuilderExtensions
         return modelBuilder;
     }
 
+    // Restricted to Ark-owned entities so the converter never silently leaks onto a
+    // consumer's own entities sharing the same DbContext. Idempotent: calling
+    // ConfigureArkEntities AND ConfigureArkPaymentEntities with the flag won't apply
+    // the converter twice (SetValueConverter on an already-converted property is a no-op,
+    // but the type-filter short-circuits the second pass cleanly anyway).
+    private static readonly IReadOnlySet<Type> ArkOwnedEntityTypes = new HashSet<Type>
+    {
+        typeof(ArkWalletEntity),
+        typeof(ArkWalletContractEntity),
+        typeof(VtxoEntity),
+        typeof(ArkIntentEntity),
+        typeof(ArkIntentVtxoEntity),
+        typeof(ArkSwapEntity),
+        typeof(ArkPaymentEntity),
+        typeof(ArkPaymentRequestEntity),
+    };
+
     private static void ApplyDateTimeOffsetTicksConversion(ModelBuilder modelBuilder)
     {
         foreach (var entityType in modelBuilder.Model.GetEntityTypes())
         {
+            if (!ArkOwnedEntityTypes.Contains(entityType.ClrType))
+                continue;
+
             foreach (var property in entityType.GetProperties())
             {
+                if (property.GetValueConverter() is not null)
+                    continue;
+
                 if (property.ClrType == typeof(DateTimeOffset))
                     property.SetValueConverter(DateTimeOffsetToTicks);
                 else if (property.ClrType == typeof(DateTimeOffset?))
--- a/NArk.Storage.EfCore/ModelBuilderExtensions.cs
+++ b/NArk.Storage.EfCore/ModelBuilderExtensions.cs
@@ -1,10 +1,33 @@
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using NArk.Storage.EfCore.Entities;
 
 namespace NArk.Storage.EfCore;
 
 public static class ModelBuilderExtensions
 {
+    // DateTimeOffset → long (UTC ticks). Applied to every DateTimeOffset property across all
+    // Ark entities. Two reasons:
+    //
+    // 1) EF Core's SQLite provider rejects `ORDER BY` on `DateTimeOffset` columns because the
+    //    default TEXT representation is not chronologically sortable across different offsets.
+    //    Storing as INTEGER (long ticks) is natively orderable on SQLite — no fallback to
+    //    client-side evaluation needed for paged queries.
+    //
+    // 2) Round-trippable and indexable across all providers (BIGINT on Postgres/MSSQL,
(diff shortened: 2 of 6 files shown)


ghost (approved):
## ✅ Approved — solid fix, two observations worth noting

**Reviewed:** all 6 changed files + full file context + interfaces + git history.

### Verdict

Clean, well-documented fix for a real SQLite limitation ([dotnet/efcore#19672](https://github.com/dotnet/efcore/issues/19672)). The `containingInputs` pagination bug fix in `EfCoreIntentStorage` is a nice bonus catch. No public API changes, no cross-repo impact — this is internal EfCore implementation only.

Not protocol-critical (no VTXO signing, forfeit paths, round lifecycle, or exit logic touched).

### Observations

**1. All providers now pay client-side sort cost, not just SQLite**

The sort+paginate is now in-memory for Postgres and MSSQL too, where `DateTimeOffset` ORDER BY works natively. For a wallet SDK where N is bounded per-wallet this is fine in practice, but worth documenting. If this becomes a hot path for multi-tenant server scenarios, consider:
- A `DateTimeOffset` → `long` (ticks) value converter for SQLite only, or
- A provider-detection branch (`db.Database.IsSqlite()`) to keep server-side sort for capable providers

Not a blocker — the pragmatic tradeoff is acknowledged in the PR description.

**2. `AsNoTracking()` inconsistency (pre-existing, not introduced here)**

`EfCoreVtxoStorage`, `EfCoreContractStorage`, and `EfCoreIntentStorage` use `.AsNoTracking().ToListAsync()`. `EfCorePaymentRequestStorage`, `EfCorePaymentStorage`, and `EfCoreSwapStorage` use `.ToListAsync()` without `AsNoTracking()`. Since these are read-only list queries, the latter three should also use `AsNoTracking()` to avoid unnecessary change-tracker overhead on the now-larger materialized sets. Consider a follow-up.

### No concerns on

- ✅ Correct sort order preserved (descending by `CreatedAt`/`SeenAt`)
- ✅ WHERE clauses still execute server-side
- ✅ `containingInputs` filter now correctly applied before `Skip`/`Take` in `GetIntents`
- ✅ Comments explain the "why" for future readers
- ✅ No interface/contract changes — downstream consumers unaffected

ghost (approved):
## ✅ Approved — much cleaner than the v1 approach, one scoping issue to consider

**Iterative review:** PR was rewritten since my last review. The client-side sort approach is gone, replaced with a `ValueConverter<DateTimeOffset, long>` opt-in. This is exactly what I suggested in observation #1 of my previous review — ORDER BY stays server-side for all providers, no in-memory materialization needed. Well done.

**Reviewed:** all 5 changed files, full file context, entity definitions, sample `WalletDbContext`, git history.

Not protocol-critical (no VTXO signing, forfeit paths, round lifecycle, or exit logic touched).

---

### Issue: `ApplyDateTimeOffsetTicksConversion` bleeds into non-Ark entities

**`ModelBuilderExtensions.cs:75-84`** — `modelBuilder.Model.GetEntityTypes()` iterates **all** entity types in the model, not just Ark ones. If a consumer's `DbContext` has `DbSet<MyEntity>` properties (auto-discovered before `OnModelCreating` runs), those entities will silently get the ticks converter applied too. The consumer didn't opt their own entities into this.

Example that breaks:
```csharp
public class AppDbContext : DbContext
{
    public DbSet<MyAuditLog> AuditLogs { get; set; } // has DateTimeOffset columns
    
    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.ConfigureArkEntities(o => o.StoreDateTimeOffsetAsTicks = true);
        // MyAuditLog.CreatedAt is now also stored as ticks — surprise!
    }
}
```

**Fix:** Maintain a set of known Ark entity CLR types and filter:
```csharp
private static readonly HashSet<Type> ArkEntityTypes = new()
{
    typeof(ArkWalletEntity), typeof(ArkWalletContractEntity),
    typeof(VtxoEntity), typeof(ArkIntentEntity), typeof(ArkIntentVtxoEntity),
    typeof(ArkSwapEntity), typeof(ArkPaymentEntity), typeof(ArkPaymentRequestEntity)
};

private static void ApplyDateTimeOffsetTicksConversion(ModelBuilder modelBuilder)
{
    foreach (var entityType in modelBuilder.Model.GetEntityTypes())
    {
        if (!ArkEntityTypes.Contains(entityType.ClrType))
            continue;
        // ... rest unchanged
    }
}
```

Not blocking approval because: (a) most consumers use a dedicated `ArkDbContext` with only Ark entities, and (b) consumers who share a DbContext can use `ConfigureConventions` instead if they want blanket conversion. But it's a correctness issue that will bite someone eventually.

### Observations (non-blocking)

**1. Sample `WalletDbContext` still uses `DateTimeOffsetToBinaryConverter`**

`samples/NArk.Wallet/NArk.Wallet.Client/Services/WalletDbContext.cs:12-15` — Uses `DateTimeOffsetToBinaryConverter` (preserves offset via `ToBinary()`), not the new `StoreDateTimeOffsetAsTicks` opt-in (drops offset via `UtcTicks`). These are semantically different converters. Consider updating the sample to dogfood the new opt-in, or at least add a comment explaining why it uses a different converter. A consumer who sees both patterns will be confused about which to use.

**2. SQL migration formula assumes SQLite ≥ 3.38.0**

`docs/articles/storage.md:140` — `julianday(SeenAt)` with timezone-offset TEXT (e.g. `2026-01-15 12:00:00+05:00`) requires SQLite 3.38.0+ for correct timezone handling. Older versions silently ignore the offset and treat the local time as UTC, producing wrong ticks. Worth adding a version note.

**3. Double `ApplyDateTimeOffsetTicksConversion` invocation**

If a consumer calls both `ConfigureArkEntities` and `ConfigureArkPaymentEntities` with the flag, the conversion loop runs twice over all entity types. Harmless (idempotent) but wasteful. Could guard with a sentinel or extract into a "call once" pattern.

### What's solid

- ✅ `UtcTicks` / `new DateTimeOffset(ticks, TimeSpan.Zero)` round-trip is correct
- ✅ Static `ValueConverter` instances are stateless and thread-safe — idiomatic EF Core
- ✅ Nullable variant handled correctly
- ✅ Trade-off (offset stripped) is well-documented in XML docs, README, and storage article
- ✅ Three targeted tests cover the fix, the trade-off, and backwards compatibility
- ✅ Default-off preserves existing consumer schemas
- ✅ No public interface changes — no cross-repo impact

ghost (approved):
## ✅ Approved — scoping fix is correct, good test coverage

**Iterative review (3rd pass):** focused on commit `76ab103d` pushed after my last review.

This commit directly addresses the entity scoping issue I flagged: `ApplyDateTimeOffsetTicksConversion` now filters to an explicit `ArkOwnedEntityTypes` set, so consumer entities sharing the same `DbContext` are never touched. The `GetValueConverter() is not null` idempotency guard is a nice bonus.

### Verified

- ✅ `ArkOwnedEntityTypes` covers all 8 Ark entity types
- ✅ `ConfigureArkEntities` + `ConfigureArkPaymentEntities` interplay is correct regardless of call order (first pass converts what's registered, second pass converts the remainder, idempotency guard skips already-converted)
- ✅ `TicksMapping_DoesNotBleedIntoConsumerEntities` test pins the scoping contract with a mixed `DbContext`
- ✅ Sample `WalletDbContext` dogfoods the opt-in, drops the old `ConfigureConventions` workaround
- ✅ SQLite ≥ 3.38.0 caveat on `julianday()` timezone handling documented
- ✅ Static `ValueConverter` instances are stateless — safe to share across `DbContext` instances

Not protocol-critical. No further concerns.

PR #95, opened 2026-05-20
fix(sync): quiet routine VtxoSynchronizationService poll logs

Kukks:
## Quiet routine VtxoSynchronizationService poll logs

The 5-second safety-net poll in `VtxoSynchronizationService.RoutinePoll` is deliberate and cheap — it bounds the arkd-indexer detection gap that motivated the v2.1.13 fix, and each tick is one short gRPC call (≈ 2-3 ms with `after=now-2min` on a small script set). The work is fine.

The **log volume** isn't. Every tick fires two `LogInformation` lines in `StartQueryLogic`:

```
info: ... StartQueryLogic: polling 2 script(s) (after=...): [...]
info: ... StartQueryLogic: poll returned 0 VTXO(s) across 2 script(s) in 3ms
```

On an idle wallet that's 24 INFO lines/minute of pure noise, drowning anything else at the default log level. Surfaced in BTCPay-plugin operator review.

### Change
- **Pre-poll "polling …" line → Debug always.** Same info is in the result line below; the upcoming arkd call isn't a productive signal on its own.
- **Post-poll result line → conditional:** `Info` only when the poll produced a VTXO (the *productive* case — a payment landed) **or** when this is the first cold-start catch-up (one-off, worth seeing). Routine ticks that find 0 drop to `Debug`.

### Non-changes
- Same gRPC call schedule and cadence (`RoutinePollInterval = 5s`, `RoutinePollLookback = 2min`).
- Same upsert path, same cursor advance, same cold-start gating.
- Operators who want the chatter back set their log level to `Debug`.

No public-API surface touched — pure logging-verbosity tune.

### Test plan
- [x] `dotnet build NArk.Core/NArk.Core.csproj` clean (0 errors).
- [ ] CI: build + unit tests + E2E.

### DIFF
--- a/NArk.Core/Services/VtxoSynchronizationService.cs
+++ b/NArk.Core/Services/VtxoSynchronizationService.cs
@@ -446,7 +446,11 @@ public class VtxoSynchronizationService : IAsyncDisposable
             var started = DateTimeOffset.UtcNow;
             try
             {
-                _logger?.LogInformation(
+                // Pre-poll line is just "we're about to call arkd" — same
+                // info is in the result line below, so keep it at Debug to
+                // avoid doubling the per-tick spam on the 5-second safety
+                // net.
+                _logger?.LogDebug(
                     "StartQueryLogic: polling {Count} script(s) (after={After}): [{Scripts}]",
                     request.Scripts.Count,
                     request.After?.ToString("O") ?? "<none>",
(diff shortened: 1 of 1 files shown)


ghost (approved):
## Review: ✅ Approve

**Scope:** Logging verbosity only — no protocol, API, or sync-logic changes.

### What I checked

1. **Full file context** (`VtxoSynchronizationService.cs`, 675 lines). The two changed log sites at lines 453 and 470-481 are inside `StartQueryLogic`, the channel-reader loop. No control flow, upsert path, cursor advancement, or cold-start gating is altered.

2. **Pre-poll log (line 453):** `LogInformation` → `LogDebug`. Correct — the same scripts/after info is repeated in the result line, so the pre-poll line was redundant at Info. No information loss.

3. **Post-poll log (lines 470-481):** Conditional level based on `found > 0 || request.IsColdStartCatchup`. The branching is sound:
   - `found > 0` → Info (a VTXO landed — the signal operators care about).
   - `IsColdStartCatchup` → Info even on 0 (one-off first-tick, good for startup diagnostics).
   - Routine 0-result ticks → Debug (the 24 lines/min noise this PR eliminates).

4. **Debug-branch log message** (line 479) hardcodes `"0"` in the format string instead of using the `{Found}` placeholder. This is fine — the branch is only entered when `found == 0`, so it's a minor optimization avoiding a boxing allocation. The structured-log template loses the `Found` property key, but since the value is always 0, no queryable signal is lost.

5. **Cold-start and cursor logic** (lines 487-508) is untouched. `_coldStartCatchupComplete`, `WriteCursorAcrossWalletsAsync`, and the `IsFullSetSnapshot` gate are identical to master.

6. **Cross-repo impact:** None. No public API, type, or interface changes. Pure internal logging.

7. **Git history:** The file has had significant churn around sync robustness (stream-miss recovery, retry schedules, per-wallet cursors). This PR correctly leaves all of that machinery alone.

### Verdict

Clean, low-risk change. Reduces operator log noise from ~24 INFO lines/min to 0 on idle wallets while preserving the productive signal (VTXO arrivals, cold-start). CI pending — approve contingent on green build.

🤖 Arkana

PR #98, opened 2026-05-21
fix(sync): demote PollScriptsForVtxos curl debug-aid log to Debug

Kukks:
## Summary
- `VtxoSynchronizationService.PollScriptsForVtxos` logs an equivalent `curl http://localhost:7070/v1/indexer/vtxos?…` line each invocation as a manual-testing aid. The line was `LogInformation`, so it fired on every 5-second safety-net poll cycle regardless of result — exactly the routine-poll noise PR #95 was meant to remove.
- Demoted to `LogDebug` and gated on `_logger.IsEnabled(LogLevel.Debug)` so the query-string formatting cost is skipped when Debug isn't enabled.

## Test plan
- [ ] At default Info log level, the `PollScriptsForVtxos: curl …` line no longer appears.
- [ ] At Debug log level, the same line appears unchanged so it stays useful as a manual-testing aid.

### DIFF
--- a/NArk.Core/Services/VtxoSynchronizationService.cs
+++ b/NArk.Core/Services/VtxoSynchronizationService.cs
@@ -612,11 +612,16 @@ public class VtxoSynchronizationService : IAsyncDisposable
         _logger?.LogInformation("PollScriptsForVtxos: querying arkd indexer for {Count} scripts (after={After}): [{Scripts}]",
             scripts.Count, after?.ToString("O") ?? "<none>", string.Join(", ", scripts));
 
-        // Log equivalent REST API URL for manual testing (substitute your arkd host:port).
-        var queryParams = string.Join("&", scripts.Select(s => $"scripts={Uri.EscapeDataString(s)}"));
-        if (after.HasValue)
-            queryParams += $"&after={after.Value.ToUnixTimeMilliseconds()}";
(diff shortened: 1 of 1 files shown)


ghost (approved):
**LGTM — approved.**

Clean, minimal change. Reviewed the full `VtxoSynchronizationService.cs` for context.

**What this does:**
- Demotes the `curl …` debug-aid log from `LogInformation` → `LogDebug` (line 623)
- Gates the `string.Join` + `Uri.EscapeDataString` formatting behind `_logger?.IsEnabled(LogLevel.Debug) == true` to skip allocation when Debug is off

**Correctness checks:**
- `_logger?.IsEnabled(LogLevel.Debug) == true` — null-safe; if `_logger` is null, `?.` returns null, `== true` is false. ✓
- Inside the guard, `_logger.LogDebug(...)` (no `?.`) is safe because the guard already proved `_logger != null`. ✓
- No behavioral change to VTXO polling, upsert, or any protocol-critical path — only the cosmetic curl log moved. ✓
- The Info-level log at line 612 (script count + after timestamp) is retained, so operators still see poll activity at default log levels. ✓
- Consistent with PR #95's intent to reduce routine-poll noise.

**Not protocol-critical** — no VTXO handling, signing, forfeit, round lifecycle, or exit path changes. Pure logging hygiene.

No cross-repo impact (no public API, type, or interface changes).

PR #99, opened 2026-05-21
perf(sync): single immediate poll per stream push, drop 750ms/3s/8s retry schedule

Kukks:
## Summary
`VtxoSynchronizationService` fanned out every arkd subscription push into three follow-up polls scheduled at +750ms / +3s / +8s. That schedule was tuned for arkd v0.9.0-rc.1's indexer commit-lag (observed range <1s to ~30s). On v0.9.5+ the indexer is essentially in sync with the subscription event, and we also still run a 5-second routine safety-net poll across every active script — so the retry schedule is now dead weight:

- Happy path waits **750ms before the first poll** even fires, so a direct-arkade transaction that was instant on arkd's side took ≥750ms before the VTXO landed in storage.
- The +3s and +8s retries are no-op `UpsertVtxo` calls (already-seen data, no `VtxosChanged` fired).

Replaced with a single immediate enqueue on each stream push. Any race the immediate poll loses against arkd's indexer is caught within ~5s by `RoutinePoll`, which already sweeps every active script with the same `after`-windowed semantics.

Removed:
- `StreamPushPollSchedule[]` (the [750ms, 3s, 8s] array)
- `FirePollSchedule` method (the per-push fan-out)

## Test plan
- [ ] `dotnet build NArk.Core` succeeds (0 errors).
- [ ] After arkd pushes an update, `UpsertVtxo: inserted ...` lands within tens of ms (vs. ~777ms before).
- [ ] If the immediate poll happens to return zero (arkd indexer lag), `RoutinePoll` catches it on the next 5s tick.
- [ ] No regression in the existing arkd-restart / stream-end-grace paths (untouched code).

### DIFF
--- a/NArk.Core/Services/VtxoSynchronizationService.cs
+++ b/NArk.Core/Services/VtxoSynchronizationService.cs
@@ -38,18 +38,6 @@ public class VtxoSynchronizationService : IAsyncDisposable
     // UpdateScriptsView full-syncs newly-added scripts with After=null.
     private static readonly TimeSpan StreamPollLookback = TimeSpan.FromMinutes(5);
 
-    // Delays between arkd's subscription push and our follow-up indexer poll(s).
-    // arkd can emit the event before the indexer query path has the VTXO visible,
-    // and we've observed the commit lag range from <1s to ~30s. Rather than pick a
-    // single tuned delay, enqueue a short retry schedule; each poll uses an `after`
-    // window so repeated fetches of unchanged data are no-ops in the upsert path.
-    private static readonly TimeSpan[] StreamPushPollSchedule =
-    [
-        TimeSpan.FromMilliseconds(750),
-        TimeSpan.FromSeconds(3),
-        TimeSpan.FromSeconds(8),
-    ];
-
     /// <summary>
     /// Internal poll work item.
     /// </summary>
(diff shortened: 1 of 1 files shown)


ghost (approved):
## ✅ Approved

Clean, low-risk perf improvement. Reviewed the full 651-line file, git history (commit `171d11a` that introduced the retry schedule), and cross-repo impact.

### What I verified

1. **No public API change.** `StreamPushPollSchedule` and `FirePollSchedule` were both private. No downstream consumers (grep'd all repos under `/root/arkana/repos/`). No conflict with open PR #98 (different region of same file).

2. **Replacement is correct.** The new immediate-enqueue code (lines 377–387) produces the same `PollRequest` shape as the old `FirePollSchedule` — same `StreamPollLookback`, same default record values (`IsFullSetSnapshot=false`, `IsColdStartCatchup=false`). The `await` on `WriteAsync` is strictly better than the old fire-and-forget pattern — no risk of accumulating concurrent timer tasks during stream-push bursts.

3. **Safety net holds.** `RoutinePoll` at 5s with 2-min lookback (lines 201–202) bounds worst-case detection latency. The old +3s/+8s retries only marginally improved the 750ms-to-5s gap — dead weight on v0.9.5+ where indexer lag is near-zero.

4. **Error handling preserved.** `OperationCanceledException` swallowed (correct for shutdown), other exceptions logged as warnings and non-fatal (RoutinePoll catches up).

5. **Not protocol-critical.** This is VTXO *detection timing*, not signing/forfeit/exit logic. Worst case of a missed immediate poll = ~5s extra latency via RoutinePoll, not fund loss.

### Minor nit (non-blocking)

**Line 79–81:** Channel comment still references "retry schedules" (`// Unbounded: retry schedules + RoutinePoll + catchup can all enqueue at once`). Now that the retry schedule is gone, this could be trimmed to just "RoutinePoll + catchup + stream-push polls". Not worth blocking on.

---
🤖 Reviewed by Arkana

PR #101, opened 2026-05-22
fix(spending): exclude unconfirmed boarding UTXOs from spendable coins

Kukks:
## Summary
- A boarding UTXO still in the mempool was being offered as a spendable coin and rendered as live/settled, even though arkd's `validateBoardingInput` rejects unconfirmed boarding inputs at settle time — so any spend built from one is doomed.
- Adds `ArkVtxo.IsUnconfirmedOnchain()` (reads a shared `Confirmed` metadata flag via the new `ArkVtxo.ConfirmedMetadataKey` const) and filters such VTXOs out of `SpendingService.GetAvailableCoins`.
- The helper is **confirmation-centric**, not boarding-specific: only boarding sync populates the flag today, but arkd-reported unrolled VTXOs will generalise for free once they carry the same flag (deliberately deferred).

## Details
- `BoardingUtxoSyncService` now writes the flag via `ArkVtxo.ConfirmedMetadataKey` instead of a string literal, so writer and readers share one constant.
- Sample wallet (`Vtxos.razor`) surfaces the state as a `PENDING` pill instead of `LIVE`.
- Docs: README boarding section + `docs/articles/spending.md` note the exclusion.

## Test plan
- [x] `dotnet build samples/NArk.Wallet/NArk.Wallet.Client` — 0 errors
- [x] `dotnet test NArk.Tests` — 397/397 pass (4 new `ArkVtxoTests` covering unconfirmed/confirmed/no-metadata/case-insensitive)

### DIFF
--- a/NArk.Abstractions/VTXOs/ArkVtxo.cs
+++ b/NArk.Abstractions/VTXOs/ArkVtxo.cs
@@ -21,6 +21,14 @@ public record ArkVtxo(
     IReadOnlyList<VtxoAsset>? Assets = null,
     Dictionary<string, string>? Metadata = null)
 {
+    /// <summary>
+    /// Metadata key carrying the on-chain confirmation state of an unrolled
+    /// (on-chain) VTXO, as <c>bool.ToString()</c> ("True"/"False"). Populated
+    /// by the boarding-UTXO sync from the explorer's confirmation status. The
+    /// writer (BoardingUtxoSyncService) and all readers share this constant.
+    /// </summary>
+    public const string ConfirmedMetadataKey = "Confirmed";
+
     public OutPoint OutPoint => new(new uint256(TransactionId), TransactionOutputIndex);
     public TxOut TxOut => new(Money.Satoshis(Amount), NBitcoin.Script.FromHex(Script));
 
@@ -56,6 +64,27 @@ public record ArkVtxo(
     public bool IsRecoverable(TimeHeight current)
     {
         return Swept || IsExpired(current) ;
+    }
+
+    /// <summary>
+    /// True when this VTXO is an on-chain output whose funding transaction is
+    /// not yet confirmed, and therefore cannot be spent or settled. Reads the
+    /// <see cref="ConfirmedMetadataKey"/> flag: a value other than "True"
+    /// (case-insensitive) means unconfirmed.
+    /// <para>
+    /// Today only boarding UTXOs carry this flag (set by the boarding-UTXO
+    /// sync from the explorer). VTXOs without the key — off-chain tree VTXOs
+    /// and arkd-reported unrolled VTXOs, which the indexer does not surface
+    /// confirmation data for — return <c>false</c>. If confirmation data is
+    /// later wired for arkd-reported unrolled VTXOs, setting the same flag
+    /// makes this (and every consumer) generalise without further changes.
+    /// </para>
+    /// </summary>
+    public bool IsUnconfirmedOnchain()
+    {
+        return Metadata is not null
+               && Metadata.TryGetValue(ConfirmedMetadataKey, out var confirmed)
+               && !string.Equals(confirmed, bool.TrueString, StringComparison.OrdinalIgnoreCase);
     }
     //
     // public bool RequiresForfeit()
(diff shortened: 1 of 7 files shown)


ghost (changes requested):
## Code Review — arkade-os/dotnet-sdk#101

**Verdict: Request changes** — the core fix is correct and well-tested, but there are two gaps where unconfirmed boarding UTXOs still leak into spend/intent paths. This is protocol-critical (VTXO handling); human sign-off required even after the gaps are addressed.

---

### ✅ What's good

1. **`ArkVtxo.IsUnconfirmedOnchain()`** (`NArk.Abstractions/VTXOs/ArkVtxo.cs:72-80`) — clean, defensive, correct. The null-metadata → `false` default is the right call for off-chain VTXOs. Case-insensitive comparison against `bool.TrueString` is solid.
2. **`ConfirmedMetadataKey` constant** — eliminates the magic-string coupling between writer and reader. Good.
3. **`SpendingService.GetAvailableCoins` filter** (`NArk.Core/Services/SpendingService.cs:160-167`) — correctly positioned before contract resolution, debug logging is useful.
4. **Tests** (`NArk.Tests/ArkVtxoTests.cs`) — four cases covering confirmed, unconfirmed, no-metadata, and case-insensitive. Solid.
5. **Docs and sample wallet** — README, spending article, and `Vtxos.razor` PENDING pill are all consistent.

---

### 🔴 P0 — Unconfirmed boarding UTXOs still reach intent generation

**`NArk.Core/Services/IntentGenerationService.cs:167-170`** — `RunIntentGenerationCycle` fetches all unspent VTXOs and feeds them directly into `intentScheduler.GetIntentsToSubmit()` without filtering `IsUnconfirmedOnchain()`. This means:

- The automatic intent loop will create intents backed by unconfirmed boarding UTXOs.
- Those intents get submitted to arkd, which rejects them at `validateBoardingInput`.
- The intent gets stuck → the 5-minute stale-intent cancellation kicks in → next cycle tries again → infinite doomed loop for every unconfirmed boarding UTXO.

**Fix:** Add the same filter at `IntentGenerationService.cs:167`:
```csharp
var walletVtxos = unspentVtxos
    .Where(v => walletContracts.Any(c => c.Script == v.Script))
    .Where(v => !recentlySpentOutpoints.Contains(...))
    .Where(v => !v.IsUnconfirmedOnchain())   // ← add this
    .ToArray();
```

---

### 🔴 P0 — `AssetManager.GetAvailableCoins` is a separate copy without the filter

**`NArk.Core/Services/AssetManager.cs:318-348`** — `AssetManager` has its own private `GetAvailableCoins` that duplicates `SpendingService.GetAvailableCoins` but does NOT filter unconfirmed VTXOs. Asset sends (`SendAsset`, `MintAsset`, `BurnAsset`) will still select unconfirmed boarding UTXOs.

**Fix:** Either:
- (a) Add the `IsUnconfirmedOnchain()` filter here too, or
- (b) Refactor both `GetAvailableCoins` copies to share a single implementation (better long-term, but out of scope for this fix).

For now, (a) is the safe minimal fix.

---

### 🟡 P1 — `Spend(walletId, inputs, outputs)` doesn't validate caller-provided inputs

**`NArk.Core/Services/SpendingService.cs:62-63`** — The manual-inputs `Spend` overload accepts `ArkCoin[] inputs` directly from the caller without checking `IsUnconfirmedOnchain()`. If a consumer builds their own coin set (bypassing `GetAvailableCoins`), they can still submit a doomed transaction.

**Suggestion:** Add a guard at the top of this method:
```csharp
var unconfirmed = inputs.Where(i => i.Vtxo?.IsUnconfirmedOnchain() == true).ToArray();
if (unconfirmed.Length > 0)
        $"Cannot spend {unconfirmed.Length} unconfirmed on-chain VTXO(s).");
```

This is lower priority since callers typically go through `GetAvailableCoins`, but it hardens the API boundary.

---

### 🟢 Nits (non-blocking)

- **`ArkVtxo.cs:68`**: The method name `IsUnconfirmedOnchain` is accurate for boarding UTXOs but slightly misleading for off-chain VTXOs that simply return `false` because they lack the key. A name like `HasPendingOnchainConfirmation` would be more precise — but the current name is acceptable and documented.

---

### Summary

The two P0 gaps mean unconfirmed boarding UTXOs still create doomed intents (via the automatic generation loop) and can be selected for asset operations. Both are straightforward fixes. Once addressed, this is a clean, well-documented bugfix.

⚠️ **Protocol-critical**: touches VTXO spend eligibility and intent generation. Requires human sign-off before merge.

🤖 Reviewed by Arkana

PR #103, opened 2026-05-22
feat(sync): update the indexer script subscription in place instead of restarting it

Kukks:
## Summary
The VTXO sync used to tear down the whole arkd `GetSubscription` stream and create a **fresh** subscription on every script-set change (`SubscribeForScripts{subscription_id:""}` → new `GetSubscription`). arkd's subscription is **mutable while streaming** (the proto says so, and the server routes events from a per-subscription channel whose topic map is mutated under lock), so this churn is avoidable — and the teardown/recreate window is exactly when a pushed event can slip through.

## Transport
Split the monolithic `GetVtxoToPollAsStream` into three composable primitives on `IClientTransport` (gRPC + REST + caching wrapper):
- `SubscribeForScriptsAsync(scripts, subscriptionId)` — create (`null` id) or **add** to an existing subscription.
- `UnsubscribeForScriptsAsync(subscriptionId, scripts)` — **remove** a subset, or all (tear down) when `null`.
- `GetVtxoSubscriptionStreamAsync(subscriptionId)` — the long-lived stream; in-place add/remove are routed onto it.

`GetVtxoToPollAsStream` is kept as a one-shot convenience **default** built on these primitives, so external callers are unaffected.

## Sync service
`VtxoSynchronizationService` now keeps **one** stream open via a supervisor loop and mutates the watched set in place:
- A contract appearing/disappearing issues `Subscribe`/`Unsubscribe` deltas — **no stream restart**.
- Reconnects on the **same** subscription id; **recreates** it if arkd reports it gone (TTL after a disconnect — detected via the "subscription … not found" error, surfaced over both gRPC and REST).
- **Tears down** the subscription when the active set is empty.
- Keeps the 5-second **fresh-derive safety-net poll** (from #102) as the backstop, so detection never depends on the stream surviving.
- A failing provider in the active-set derivation is still isolated.

Design decisions confirmed with the maintainer: reconnect-same-id (recreate only on not-found), keep the poll backstop, tear down on empty.

## Test plan
- [x] `dotnet test NArk.Tests` — **404/404** green
- [x] 13 `VtxoSynchronizationService` tests incl. in-place add (asserts the stream is opened exactly once), in-place remove, GC recreate, empty teardown, poll self-heal, provider isolation, cold-start cursor
- [ ] CI green on this PR

### DIFF
--- a/NArk.Core/Services/VtxoSynchronizationService.cs
+++ b/NArk.Core/Services/VtxoSynchronizationService.cs
@@ -21,8 +21,19 @@ public class VtxoSynchronizationService : IAsyncDisposable
     private readonly CancellationTokenSource _shutdownCts = new();
     private Task? _queryTask;
 
-    private CancellationTokenSource? _restartCts;
+    // The long-lived stream supervisor task. It keeps exactly one GetSubscription
+    // stream open and updates the watched set IN PLACE (Subscribe/Unsubscribe) rather
+    // than tearing the stream down on every script change.
     private Task? _streamTask;
+    // Cancelled to interrupt the supervisor's current stream so it re-reads the
+    // subscription state (on recreate or teardown). A fresh instance per generation.
+    private CancellationTokenSource? _streamGenerationCts;
+    // Released when a subscription is (re)created from the idle/no-subscription state,
+    // to wake the supervisor out of its idle wait.
+    private readonly SemaphoreSlim _streamWakeup = new(0);
+    // arkd subscription id for the current in-place subscription; null when there is
+    // none (empty active set, or torn down).
+    private string? _subscriptionId;
 
     /// <summary>
     /// The script set the subscription stream is currently subscribed to.
@@ -211,6 +222,8 @@ public class VtxoSynchronizationService : IAsyncDisposable
     // Tunable for tests via the internal init property.
     internal TimeSpan RoutinePollInterval { get; init; } = TimeSpan.FromSeconds(5);
     private static readonly TimeSpan RoutinePollLookback = TimeSpan.FromMinutes(2);
+    // Backoff before reconnecting the subscription stream after a transient fault.
+    private static readonly TimeSpan StreamReconnectDelay = TimeSpan.FromSeconds(1);
     private Task? _routinePollTask;
 
     public async Task StartAsync(CancellationToken cancellationToken)
@@ -219,6 +232,9 @@ public class VtxoSynchronizationService : IAsyncDisposable
         var multiToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _shutdownCts.Token);
         _queryTask = StartQueryLogic(multiToken.Token);
         _routinePollTask = RoutinePoll(multiToken.Token);
+        // The supervisor starts idle and wakes once UpdateScriptsView creates the
+        // initial subscription below.
+        _streamTask = RunStreamSupervisorAsync(multiToken.Token);
         await UpdateScriptsView(multiToken.Token);
     }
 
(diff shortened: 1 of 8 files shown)


ghost (approved):
## Code Review — PR #103: feat(sync): in-place subscription update

**Reviewer: Arkana (automated, aggressive)**
**Verdict: APPROVE** — well-structured change, no protocol-critical or correctness issues found.

---

### Summary

This PR replaces the tear-down-and-recreate pattern for arkd script subscriptions with an in-place update model. The key improvement: the old code had a window during stream restart where pushed events could be lost. The new code keeps one long-lived stream and mutates the watched set via `Subscribe`/`Unsubscribe` deltas. The 5-second safety-net poll is preserved as the backstop — good defensive design.

### What I checked

- Full diff (8 files, +535/−202)
- `DisposeAsync` shutdown path for proper task awaiting and resource cleanup
- `_viewSyncLock` acquisition patterns for deadlocks
- Cross-repo impact (ts-sdk, go-sdk, rust-sdk) — **none**, `IClientTransport` is SDK-internal
- All 4 `IClientTransport` implementations (gRPC, REST, Caching, test mock) updated consistently
- `GetVtxoToPollAsStream` retained as default interface method — backward compatible
- No conflicting open PRs on this repo

### Findings

**1. `IsSubscriptionNotFound` string matching is fragile** (VtxoSynchronizationService.cs)
```csharp
private static bool IsSubscriptionNotFound(Exception ex)
    => ex.Message.Contains("not found", StringComparison.OrdinalIgnoreCase)
       || (ex.InnerException?.Message.Contains("not found", StringComparison.OrdinalIgnoreCase) == true);
```
Any exception with "not found" in the message triggers the recreate path. This is safe (recreating a valid subscription is a no-op minus a round-trip), but could mask real errors behind a recreation attempt. Consider matching on the subscription id in the message too, e.g. `$"subscription {id} not found"`, or introduce a typed exception / error code if arkd ever stabilizes this error surface. **Non-blocking** — the fallback behavior is correct.

**2. `SignalStreamWakeup` has a benign TOCTOU race** (VtxoSynchronizationService.cs)
```csharp
private void SignalStreamWakeup()
{
    if (_streamWakeup.CurrentCount == 0)
        _streamWakeup.Release();
}
```
Two concurrent callers could both see `CurrentCount == 0` and both `Release()`, pushing the semaphore to 2. This just causes a harmless extra supervisor loop iteration (it re-reads `_subscriptionId`, finds no change, goes idle). **Non-blocking** — spurious wakeup is fine.

**3. `_viewSyncLock` held during network I/O** (VtxoSynchronizationService.cs, UpdateScriptsView + ReassertSubscriptionAsync)
Both methods hold the lock while calling `SubscribeForScriptsAsync`/`UnsubscribeForScriptsAsync` (network round-trips to arkd). If arkd is slow, all callers needing `_viewSyncLock` block. This is **intentional and necessary** — releasing the lock before the network call would create a race between the subscription state mutation and a concurrent `OnActiveScriptsChanged`. The old code had the same pattern (held lock during `await _streamTask` which also did network I/O). **Non-blocking**.

**4. Old recursive deadlock risk eliminated** ✅
The old `StartStreamLogic` error handler called `UpdateScriptsView` recursively, which could deadlock on `_viewSyncLock` (SemaphoreSlim is not reentrant). The new supervisor pattern (`RunStreamSupervisorAsync` → `ReassertSubscriptionAsync`) acquires the lock independently with no recursion. Good fix.

**5. Minor test coverage gap**
`ReassertSubscriptionAsync` (stream drops → reassert same id → arkd says "not found" → recreate) is not directly tested. The `SubscriptionGarbageCollected_Recreates` test covers the `UpdateScriptsView` path (in-place add fails with "not found"), but not the supervisor's reconnect path. Consider adding a test that simulates a stream drop followed by a "not found" during reassertion. **Non-blocking**.

**6. REST vs gRPC parity for null scripts in Unsubscribe**
- gRPC: sends empty repeated field (no `AddRange`)
- REST: sends `scripts: []` (empty array)
Both should be interpreted identically by arkd. Verified consistent. ✅

### Protocol assessment

This change touches VTXO **detection/sync**, not transaction signing, forfeit paths, round lifecycle, or exit paths. A bug here could delay VTXO detection (worst case ~5s until safety-net poll catches it), but cannot cause fund loss. The safety-net poll backstop is preserved. **Not protocol-critical**.

### Verdict

Clean design. The supervisor pattern is a strict improvement over tear-down-and-recreate. All implementations updated consistently. Tests cover the key scenarios. The three findings above are all non-blocking.

---
🤖 Reviewed by Arkana

ghost (approved):
## Code Review — PR #103: feat(sync): in-place subscription update

**Reviewer: Arkana (automated, aggressive)**
**Verdict: APPROVE** — well-structured change, no protocol-critical or correctness issues found.

---

### Summary

This PR replaces the tear-down-and-recreate pattern for arkd script subscriptions with an in-place update model. The key improvement: the old code had a window during stream restart where pushed events could be lost. The new code keeps one long-lived stream and mutates the watched set via Subscribe/Unsubscribe deltas. The 5-second safety-net poll is preserved as the backstop — good defensive design.

### What I checked

- Full diff (8 files, +535/−202)
- DisposeAsync shutdown path for proper task awaiting and resource cleanup
- _viewSyncLock acquisition patterns for deadlocks
- Cross-repo impact (ts-sdk, go-sdk, rust-sdk) — **none**, IClientTransport is SDK-internal
- All 4 IClientTransport implementations (gRPC, REST, Caching, test mock) updated consistently
- GetVtxoToPollAsStream retained as default interface method — backward compatible
- No conflicting open PRs on this repo

### Findings

**1. IsSubscriptionNotFound string matching is fragile** (VtxoSynchronizationService.cs)

Any exception with "not found" in the message triggers the recreate path. This is safe (recreating a valid subscription is a no-op minus a round-trip), but could mask real errors behind a recreation attempt. Consider matching on the subscription id in the message too, or introduce a typed exception / error code if arkd ever stabilizes this error surface. **Non-blocking** — the fallback behavior is correct.

**2. SignalStreamWakeup has a benign TOCTOU race** (VtxoSynchronizationService.cs)

Two concurrent callers could both see CurrentCount == 0 and both Release(), pushing the semaphore to 2. This just causes a harmless extra supervisor loop iteration (re-reads _subscriptionId, finds no change, goes idle). **Non-blocking** — spurious wakeup is fine.

**3. _viewSyncLock held during network I/O** (VtxoSynchronizationService.cs, UpdateScriptsView + ReassertSubscriptionAsync)

Both methods hold the lock while calling SubscribeForScriptsAsync/UnsubscribeForScriptsAsync (network round-trips to arkd). If arkd is slow, all callers needing _viewSyncLock block. This is **intentional and necessary** — releasing the lock before the network call would create a race between the subscription state mutation and a concurrent OnActiveScriptsChanged. The old code had the same pattern. **Non-blocking**.

**4. Old recursive deadlock risk eliminated** ✅

The old StartStreamLogic error handler called UpdateScriptsView recursively, which could deadlock on _viewSyncLock (SemaphoreSlim is not reentrant). The new supervisor pattern (RunStreamSupervisorAsync → ReassertSubscriptionAsync) acquires the lock independently with no recursion. Good fix.

**5. Minor test coverage gap**

ReassertSubscriptionAsync (stream drops → reassert same id → arkd says "not found" → recreate) is not directly tested. The SubscriptionGarbageCollected_Recreates test covers the UpdateScriptsView path, but not the supervisor's reconnect path. Consider adding a test that simulates a stream drop followed by a "not found" during reassertion. **Non-blocking**.

**6. REST vs gRPC parity for null scripts in Unsubscribe** ✅

Both send empty/absent scripts lists. Verified consistent.

### Protocol assessment

This change touches VTXO **detection/sync**, not transaction signing, forfeit paths, round lifecycle, or exit paths. A bug here could delay VTXO detection (worst case ~5s until safety-net poll catches it), but cannot cause fund loss. The safety-net poll backstop is preserved. **Not protocol-critical**.

### Verdict

Clean design. The supervisor pattern is a strict improvement over tear-down-and-recreate. All implementations updated consistently. Tests cover the key scenarios. The three findings above are all non-blocking.

---
🤖 Reviewed by Arkana

PR #104, opened 2026-05-27
feat(recovery): unified wallet recovery + legacy deprecated-signer discovery

Kukks:
## Unified wallet recovery (all wallet types) + legacy-signer discovery

Adds a wallet-type-agnostic recovery entry point and closes a gap where funds
locked under **legacy script variants** were never rediscovered.

### Legacy-signer discovery (the core fix)
`IndexerVtxoDiscoveryProvider` previously derived only `ArkPaymentContract` against
the **current** signer. Server-key rotation leaves earlier funds under a different
script (a different server key), so recovery silently missed them — unlike the
canonical `arkade-os/ts-sdk` restore, which derives across signers + script variants.

Now, per index, it derives the payment contract (and the delegate contract for each
configured `RecoveryDelegateConfig.Delegates`) against the **current signer AND every
`DeprecatedSigner`**, probes all candidate scripts in one indexer query, and returns
every match. Exit delay is invariant; only the server key varies. Adds
`KeyExtensions.ToOutputDescriptor(ECXOnlyPubKey)` → `tr(<32-byte x-only>)` to rebuild a
descriptor for the x-only deprecated signers the SDK stores (the vendored parser
already supports x-only `tr()`).

### Unified `IWalletRecoveryService` (NArk.Swaps)
A single `RecoverAsync(walletId)` that dispatches by wallet type:
- **HD** → `HdWalletRecoveryService.ScanAsync` (gap-limit index scan across signers; restores boltz swaps in-line via the discovery providers)
- **SingleKey** → ensure the deterministic contract exists, then `RestoreSwaps` for its descriptor

then, for both, finalizes pending Arkade transactions, resyncs funds for the recovered
offchain scripts, and audits swap recoverability. Returns a `WalletRecoveryReport`.
Lives in NArk.Swaps (the layer that can see both Core recovery + swap services);
registered by `AddArkSwapServices`.

### Tests
- **Unit (NArk.Tests):** `IndexerVtxoDiscoveryProviderTests` — proves a VTXO under a
  deprecated signer is discovered (and returns the legacy contract). Full suite **407/407** green.
- **E2E (NArk.Tests.End2End, Swaps):** `WalletRecoveryTests` — real-data round-trip on
  the production wallet stack: fund HD wallet → reverse boltz swap → wipe storage →
  `RecoverAsync` → assert contracts + index + funds + swap restored. Compiles locally;
  runtime exercised by the `e2e` CI job (arkd + boltz).

### Docs
README "Wallet Recovery" usage section + XML docs on all new public APIs.

> Consumed by the BTCPay Arkade plugin (separate PR): wallet import triggers
> `RecoverAsync` in the background, plus a manual "Rescan" action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,20 +37,13 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 9.0.x
-      - name: Setup Go
-        uses: actions/setup-go@v5
+      - name: Setup Node
+        uses: actions/setup-node@v4
         with:
-          go-version: '1.23'
-      - name: Cache nigiri binary
-        uses: actions/cache@v4
-        with:
-          path: regtest/_build
-          key: nigiri-${{ runner.os }}-${{ hashFiles('regtest/.env.defaults', '.env.regtest') }}
+          node-version: '20'
       - name: Start regtest infrastructure
-        run: |
-          chmod +x regtest/*.sh regtest/helpers/*.sh
-          ./regtest/start-env.sh
-        timeout-minutes: 10
+        run: node regtest/regtest.mjs start --profile boltz,delegate
+        timeout-minutes: 20
       - name: Restore and build
         run: dotnet build
       - name: Run E2E tests
@@ -59,7 +52,7 @@ jobs:
       - name: Collect docker logs on failure
         if: failure()
         run: |
-          docker logs ark > /tmp/arkd-logs.txt 2>&1 || true
+          docker logs arkd > /tmp/arkd-logs.txt 2>&1 || true
           docker logs boltz > /tmp/boltz-logs.txt 2>&1 || true
           docker logs bitcoin > /tmp/bitcoin-logs.txt 2>&1 || true
           docker logs lnd > /tmp/lnd-logs.txt 2>&1 || true
@@ -73,9 +66,7 @@ jobs:
           path: /tmp/*-logs.txt
       - name: Stop infrastructure
         if: always()
-        run: |
-          chmod +x regtest/*.sh regtest/helpers/*.sh 2>/dev/null
-          ./regtest/clean-env.sh || true
+        run: node regtest/regtest.mjs clean || true
 
   publish:
     name: Publish NuGet Packages
--- a/NArk.Abstractions/Extensions/KeyExtensions.cs
+++ b/NArk.Abstractions/Extensions/KeyExtensions.cs
@@ -45,6 +45,14 @@ public static class KeyExtensions
         return ParseCached($"tr({str})", network);
     }
 
+    /// <summary>
+    /// Builds a <c>tr(&lt;32-byte x-only&gt;)</c> output descriptor from an x-only
+    /// public key (lowercase hex). Used to reconstruct a descriptor for a server
+    /// signer the SDK stores x-only — e.g. a deprecated signer during recovery.
+    /// </summary>
+    public static OutputDescriptor ToOutputDescriptor(this ECXOnlyPubKey pubKey, Network network)
+        => ParseOutputDescriptor(Encoders.Hex.EncodeData(pubKey.ToBytes()), network);
+
     // NBitcoin's OutputDescriptor.Parse is observed at ~500ms-1s per call on
     // wildcard taproot descriptors (BIP-380 lexer + key-origin + checksum
     // validation in a hot codepath). The same descriptor strings (server,
--- a/NArk.Abstractions/Recovery/RecoveryDelegateConfig.cs
+++ b/NArk.Abstractions/Recovery/RecoveryDelegateConfig.cs
@@ -1,0 +1,16 @@
+using NBitcoin.Scripting;
+
+namespace NArk.Abstractions.Recovery;
+
+/// <summary>
+/// Optional delegate (auto-renewal) public-key descriptors a wallet may have used.
+/// When non-empty, recovery also derives the delegate-vtxo variant per index/signer
+/// — mirroring the canonical <c>arkade-os/ts-sdk</c> restore, which derives both
+/// <c>DefaultVtxo</c> and <c>DelegateVtxo</c> scripts — so funds locked under a
+/// delegate script are discovered too. Empty by default (no delegation configured).
+/// </summary>
+public sealed class RecoveryDelegateConfig
+{
+    /// <summary>Delegate key descriptors to additionally probe during recovery.</summary>
+    public IReadOnlyList<OutputDescriptor> Delegates { get; init; } = [];
+}
--- a/NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs
+++ b/NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs
@@ -1,5 +1,6 @@
 using Microsoft.Extensions.Logging;
 using NArk.Abstractions.Contracts;
+using NArk.Abstractions.Extensions;
 using NArk.Abstractions.Recovery;
 using NArk.Abstractions.Wallets;
 using NArk.Core.Contracts;
@@ -9,22 +10,39 @@ using NBitcoin.Scripting;
 namespace NArk.Core.Recovery;
 
 /// <summary>
-/// Discovery provider that asks arkd's indexer whether the
-/// <see cref="ArkPaymentContract"/> derived from a given HD index has any
-/// VTXO recorded against it (spent or unspent). A single VTXO at any state
-/// is sufficient evidence the index was used.
+/// Discovery provider that asks arkd's indexer whether any VTXO (spent or unspent)
+/// is recorded against the contracts derivable from a given HD index. A single VTXO
+/// at any state is sufficient evidence the index was used.
+/// <para>
+/// To recover funds locked under <b>legacy</b> script formats, this derives and
+/// probes more than the current default contract — mirroring the canonical
+/// <c>arkade-os/ts-sdk</c> restore. For each index it builds, for <b>every</b>
+/// signer in <c>{ current SignerKey } ∪ DeprecatedSigners</c> (server-key rotation
+/// leaves old funds under a different script):
+/// </para>
+/// <list type="bullet">
+///   <item><see cref="ArkPaymentContract"/> (the default VTXO script), and</item>
+///   <item><see cref="ArkDelegateContract"/> (the delegate VTXO script) for each
+///   configured delegate descriptor (<see cref="RecoveryDelegateConfig"/>), if any.</item>
+/// </list>
+/// The exit delay is invariant across signers; only the server key changes. Every
+/// candidate whose script the indexer reports a VTXO for is returned so the
+/// orchestrator persists it.
 /// </summary>
 public class IndexerVtxoDiscoveryProvider(
     IClientTransport clientTransport,
+    RecoveryDelegateConfig? delegateConfig = null,
     ILogger<IndexerVtxoDiscoveryProvider>? logger = null) : IContractDiscoveryProvider
 {
-    // ArkServerInfo is invariant for the wallet-recovery use case (signer key,
+    // ArkServerInfo is invariant for the wallet-recovery use case (signer keys,
     // exit delays, network — all server-side config that doesn't change between
     // probes). Cache the fetch on first use and reuse for every subsequent
     // index — saves N round-trips per scan when a wallet has dozens of indices.
     private readonly Lazy<Task<ArkServerInfo>> _serverInfo = new(
         () => clientTransport.GetServerInfoAsync(),
         LazyThreadSafetyMode.ExecutionAndPublication);
+
+    private readonly IReadOnlyList<OutputDescriptor> _delegates = delegateConfig?.Delegates ?? [];
 
     /// <inheritdoc />
     public string Name => "indexer";
@@ -37,19 +55,54 @@ public class IndexerVtxoDiscoveryProvider(
         CancellationToken cancellationToken = default)
     {
         var serverInfo = await _serverInfo.Value.WaitAsync(cancellationToken);
-        var contract = new ArkPaymentContract(serverInfo.SignerKey, serverInfo.UnilateralExit, userDescriptor);
-        var script = contract.GetScriptPubKey().ToHex();
 
-        await foreach (var _ in clientTransport.GetVtxoByScriptsAsSnapshot(
-                           new HashSet<string> { script }, cancellationToken))
+        // Build the full candidate set: default (+ delegate) contracts against the
+        // current signer and every deprecated signer. Keyed by scriptPubKey hex so
+        // a returned VTXO maps back to the exact contract that owns it.
+        var byScript = BuildCandidates(serverInfo, userDescriptor);
+
+        var matched = new Dictionary<string, ArkContract>(StringComparer.OrdinalIgnoreCase);
+        await foreach (var vtxo in clientTransport.GetVtxoByScriptsAsSnapshot(
+                           byScript.Keys.ToHashSet(), cancellationToken))
         {
-            // Any VTXO is enough — return immediately to avoid streaming the whole script's history.
-            logger?.LogDebug(
-                "IndexerVtxoDiscoveryProvider: hit at index {Index} on script {Script}",
-                index, script);
-            return new DiscoveryResult(true, [contract]);
+            if (byScript.TryGetValue(vtxo.Script, out var contract) && matched.TryAdd(vtxo.Script, contract))
+            {
+                logger?.LogDebug(
+                    "IndexerVtxoDiscoveryProvider: hit at index {Index} on script {Script} ({Type})",
+                    index, vtxo.Script, contract.Type);
+                // Stop early once every candidate has a hit — nothing more to learn.
+                if (matched.Count == byScript.Count)
+                    break;
+            }
         }
 
-        return DiscoveryResult.NotFound;
+        return matched.Count == 0
+            ? DiscoveryResult.NotFound
+            : new DiscoveryResult(true, matched.Values.ToList());
     }
+
+    private Dictionary<string, ArkContract> BuildCandidates(
+        ArkServerInfo serverInfo, OutputDescriptor userDescriptor)
+    {
+        // Current signer first, then every deprecated signer (server-key rotation).
+        // DeprecatedSigners maps key -> cutoff timestamp; only the key matters here,
+        // the exit delay is the server-wide UnilateralExit.
+        var signers = new List<OutputDescriptor> { serverInfo.SignerKey };
+        foreach (var deprecated in serverInfo.DeprecatedSigners.Keys)
+            signers.Add(deprecated.ToOutputDescriptor(serverInfo.Network));
+
+        var byScript = new Dictionary<string, ArkContract>(StringComparer.OrdinalIgnoreCase);
+        foreach (var signer in signers)
+        {
+            AddCandidate(byScript, new ArkPaymentContract(signer, serverInfo.UnilateralExit, userDescriptor));
+            foreach (var delegateDescriptor in _delegates)
+                AddCandidate(byScript, new ArkDelegateContract(
+                    signer, serverInfo.UnilateralExit, userDescriptor, delegateDescriptor));
+        }
+
+        return byScript;
+    }
+
+    private static void AddCandidate(Dictionary<string, ArkContract> byScript, ArkContract contract)
+        => byScript.TryAdd(contract.GetScriptPubKey().ToHex(), contract);
 }
--- a/NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs
+++ b/NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs
@@ -36,13 +36,31 @@ public class IndexerVtxoDiscoveryProvider(
 {
     // ArkServerInfo is invariant for the wallet-recovery use case (signer keys,
     // exit delays, network — all server-side config that doesn't change between
-    // probes). Cache the fetch on first use and reuse for every subsequent
-    // index — saves N round-trips per scan when a wallet has dozens of indices.
-    private readonly Lazy<Task<ArkServerInfo>> _serverInfo = new(
-        () => clientTransport.GetServerInfoAsync(),
-        LazyThreadSafetyMode.ExecutionAndPublication);
+    // probes). Cache the successful RESULT (not a Task) and reuse it for every
+    // subsequent index — saves N round-trips per scan. We deliberately do NOT
+    // cache a Lazy<Task>: a Lazy would permanently publish a *faulted* task if
+    // the first fetch fails (e.g. arkd restarts mid-scan), poisoning every later
+    // probe into a silent zero-result recovery. Caching the result lets a
+    // transient failure be retried on the next index.
+    private ArkServerInfo? _serverInfo;
+    private readonly SemaphoreSlim _serverInfoLock = new(1, 1);
 
     private readonly IReadOnlyList<OutputDescriptor> _delegates = delegateConfig?.Delegates ?? [];
+
+    private async Task<ArkServerInfo> GetServerInfoAsync(CancellationToken cancellationToken)
+    {
+        if (_serverInfo is { } cached)
+            return cached;
+        await _serverInfoLock.WaitAsync(cancellationToken);
+        try
+        {
+            return _serverInfo ??= await clientTransport.GetServerInfoAsync(cancellationToken);
+        }
+        finally
+        {
+            _serverInfoLock.Release();
+        }
+    }
 
     /// <inheritdoc />
     public string Name => "indexer";
@@ -54,7 +72,7 @@ public class IndexerVtxoDiscoveryProvider(
         int index,
         CancellationToken cancellationToken = default)
     {
-        var serverInfo = await _serverInfo.Value.WaitAsync(cancellationToken);
+        var serverInfo = await GetServerInfoAsync(cancellationToken);
 
         // Build the full candidate set: default (+ delegate) contracts against the
         // current signer and every deprecated signer. Keyed by scriptPubKey hex so
--- a/NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs
+++ b/NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs
@@ -5,6 +5,7 @@ using NArk.Abstractions.Recovery;
 using NArk.Abstractions.Wallets;
 using NArk.Core.Contracts;
 using NArk.Core.Transport;
+using NBitcoin;
 using NBitcoin.Scripting;
 
 namespace NArk.Core.Recovery;
@@ -25,8 +26,16 @@ namespace NArk.Core.Recovery;
 ///   <item><see cref="ArkDelegateContract"/> (the delegate VTXO script) for each
 ///   configured delegate descriptor (<see cref="RecoveryDelegateConfig"/>), if any.</item>
 /// </list>
-/// The exit delay is invariant across signers; only the server key changes. Every
-/// candidate whose script the indexer reports a VTXO for is returned so the
+/// <para>
+/// On mainnet the candidate set also pairs each signer with the historical 7-day
+/// unilateral-exit delay (605184s) alongside the arkd-advertised one — matching
+/// ts-sdk's <c>MAINNET_UNILATERAL_EXIT_DELAY</c> fallback. arkd only advertises
+/// the CURRENT delay; without this, wallets that minted VTXOs while mainnet still
+/// ran the original delay would silently fail discovery after the operator
+/// shortened it. We keep one hardcoded fallback rather than scanning an unbounded
+/// history because arkd does not record deprecated delays.
+/// </para>
+/// Every candidate whose script the indexer reports a VTXO for is returned so the
 /// orchestrator persists it.
 /// </summary>
 public class IndexerVtxoDiscoveryProvider(
@@ -46,6 +55,13 @@ public class IndexerVtxoDiscoveryProvider(
     private readonly SemaphoreSlim _serverInfoLock = new(1, 1);
 
     private readonly IReadOnlyList<OutputDescriptor> _delegates = delegateConfig?.Delegates ?? [];
+
+    // Mainnet's original unilateral-exit delay (~7 days in seconds). Kept as a
+    // single hardcoded fallback paired with the arkd-advertised delay so wallets
+    // that minted VTXOs before the operator shortened the delay still discover
+    // them. Matches MAINNET_UNILATERAL_EXIT_DELAY in arkade-os/ts-sdk's restore.
+    private static readonly Sequence MainnetLegacyUnilateralExit =
+        new(TimeSpan.FromSeconds(605184));
 
     private async Task<ArkServerInfo> GetServerInfoAsync(CancellationToken cancellationToken)
     {
@@ -74,9 +90,11 @@ public class IndexerVtxoDiscoveryProvider(
     {
         var serverInfo = await GetServerInfoAsync(cancellationToken);
 
-        // Build the full candidate set: default (+ delegate) contracts against the
-        // current signer and every deprecated signer. Keyed by scriptPubKey hex so
-        // a returned VTXO maps back to the exact contract that owns it.
+        // Build the full candidate set: default (+ delegate) contracts across the
+        // cross-product of { current signer ∪ deprecated signers } × { current
+        // exit delay ∪ mainnet legacy delay (mainnet only) }. Keyed by
+        // scriptPubKey hex so a returned VTXO maps back to the exact contract
+        // that owns it.
         var byScript = BuildCandidates(serverInfo, userDescriptor);
 
         var matched = new Dictionary<string, ArkContract>(StringComparer.OrdinalIgnoreCase);
@@ -103,19 +121,28 @@ public class IndexerVtxoDiscoveryProvider(
         ArkServerInfo serverInfo, OutputDescriptor userDescriptor)
     {
         // Current signer first, then every deprecated signer (server-key rotation).
-        // DeprecatedSigners maps key -> cutoff timestamp; only the key matters here,
-        // the exit delay is the server-wide UnilateralExit.
+        // DeprecatedSigners maps key -> cutoff timestamp; only the key matters here.
         var signers = new List<OutputDescriptor> { serverInfo.SignerKey };
         foreach (var deprecated in serverInfo.DeprecatedSigners.Keys)
             signers.Add(deprecated.ToOutputDescriptor(serverInfo.Network));
 
+        // Current exit delay + the hardcoded mainnet legacy delay (deduped). On
+        // testnets arkd has always advertised the network's intended delay, so
+        // the legacy probe would only widen the candidate set without ever
+        // hitting — keep it mainnet-only to match ts-sdk and avoid wasting
+        // indexer round-trips on every regtest scan.
+        var exitDelays = new HashSet<Sequence> { serverInfo.UnilateralExit };
+        if (serverInfo.Network.ChainName == ChainName.Mainnet)
+            exitDelays.Add(MainnetLegacyUnilateralExit);
+
         var byScript = new Dictionary<string, ArkContract>(StringComparer.OrdinalIgnoreCase);
         foreach (var signer in signers)
+        foreach (var exitDelay in exitDelays)
         {
-            AddCandidate(byScript, new ArkPaymentContract(signer, serverInfo.UnilateralExit, userDescriptor));
+            AddCandidate(byScript, new ArkPaymentContract(signer, exitDelay, userDescriptor));
             foreach (var delegateDescriptor in _delegates)
                 AddCandidate(byScript, new ArkDelegateContract(
-                    signer, serverInfo.UnilateralExit, userDescriptor, delegateDescriptor));
+                    signer, exitDelay, userDescriptor, delegateDescriptor));
         }
 
         return byScript;
--- a/NArk.Swaps/Hosting/SwapServiceCollectionExtensions.cs
+++ b/NArk.Swaps/Hosting/SwapServiceCollectionExtensions.cs
@@ -26,6 +26,9 @@ public static class SwapServiceCollectionExtensions
     {
         // Core services (provider-agnostic)
         services.AddSingleton<SwapsManagementService>();
+        // Unified, wallet-type-agnostic recovery facade. Lives here because it
+        // composes both the Core recovery services and the swap services.
+        services.AddSingleton<IWalletRecoveryService, WalletRecoveryService>();
         services.AddSingleton<ISweepPolicy, SwapSweepPolicy>();
         services.AddSingleton<IContractTransformer, VHTLCContractTransformer>();
         services.AddHostedService<NArk.Swaps.Hosting.SwapHostedLifecycle>();
--- a/NArk.Swaps/Recovery/IWalletRecoveryService.cs
+++ b/NArk.Swaps/Recovery/IWalletRecoveryService.cs
@@ -1,0 +1,27 @@
+using NArk.Abstractions.Recovery;
+
+namespace NArk.Swaps.Recovery;
+
+/// <summary>
+/// Wallet-type-agnostic recovery entry point. Given a wallet id, rebuilds local
+/// state from on-chain / indexer / boltz sources: contracts (incl. legacy script
+/// variants under deprecated server signers), the HD derivation index, funds
+/// (VTXOs), boltz swap data, and any in-flight Ark transactions.
+/// <para>
+/// Dispatches by wallet type: HD wallets get a gap-limit index scan (which also
+/// restores boltz swaps via the discovery providers); SingleKey wallets — whose
+/// contract set is fixed by their one key — re-derive that contract and restore
+/// swaps directly. Both then finalize pending transactions and sync funds.
+/// </para>
+/// </summary>
+public interface IWalletRecoveryService
+{
+    /// <summary>Recover everything reconstructable for <paramref name="walletId"/>.</summary>
+    /// <param name="walletId">The wallet to recover (must already exist in storage).</param>
+    /// <param name="options">HD scan tuning (gap limit, max index). Ignored for SingleKey.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    Task<WalletRecoveryReport> RecoverAsync(
+        string walletId,
+        RecoveryOptions? options = null,
+        CancellationToken cancellationToken = default);
+}
--- a/NArk.Swaps/Recovery/WalletRecoveryReport.cs
+++ b/NArk.Swaps/Recovery/WalletRecoveryReport.cs
@@ -1,0 +1,35 @@
+using NArk.Abstractions.Recovery;
+using NArk.Abstractions.Wallets;
+using NArk.Swaps.Models;
+
+namespace NArk.Swaps.Recovery;
+
+/// <summary>
+/// Outcome of a unified <see cref="IWalletRecoveryService.RecoverAsync"/> run.
+/// </summary>
+/// <param name="WalletType">The recovered wallet's type (HD vs SingleKey).</param>
+/// <param name="HdScan">
+/// The HD index-scan report (contracts + highest used index), or <c>null</c> for
+/// a SingleKey wallet (whose contract set is fixed by its single key — no scan).
+/// </param>
+/// <param name="ContractsRecovered">Total contracts persisted for the wallet after recovery.</param>
+/// <param name="RestoredSwaps">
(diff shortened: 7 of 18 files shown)


ghost (changes requested):
## 🔍 Arkana Code Review — arkade-os/dotnet-sdk#104

**Verdict: Request changes** — two concrete recovery-completeness gaps that will silently miss funds, plus this is protocol-critical (VTXO discovery / wallet recovery) and **requires human sign-off** regardless.

Overall the PR is well-structured, well-documented, and solves a real gap (legacy deprecated-signer discovery). The unified `IWalletRecoveryService` facade is clean. But two paths through the new code can't recover funds they should, and one is a script-mismatch bug.

---

### 🚨 P0 — Delegate contract recovery ignores `cltvLocktime` (script mismatch → missed funds)

**`NArk.Core/Recovery/IndexerVtxoDiscoveryProvider.cs:99-100`**

```csharp
AddCandidate(byScript, new ArkDelegateContract(
    signer, serverInfo.UnilateralExit, userDescriptor, delegateDescriptor));
```

This always passes `null` for `cltvLocktime` (the 5th param defaults to `null`). But in production, `DelegatingAddressProvider` (`NArk.Core/Wallet/DelegatingAddressProvider.cs:46-51`) creates delegate contracts with a caller-supplied `cltvLocktime` that **changes the tapscript tree and therefore the scriptPubKey**. A delegate VTXO originally locked with a non-null locktime will have a different script than the one recovery derives → the indexer query will never match it → **funds silently lost**.

**Fix**: `RecoveryDelegateConfig` should carry the `LockTime?` alongside each delegate descriptor (or a list of locktime variants to probe). `BuildCandidates` must iterate over them:

```csharp
// Sketch:
public sealed class RecoveryDelegateConfig
{
    public IReadOnlyList<(OutputDescriptor Descriptor, LockTime? CltvLocktime)> Delegates { get; init; } = [];
}
```

---

### ⚠️ P1 — SingleKey wallet recovery doesn't probe deprecated signers

**`NArk.Swaps/Recovery/WalletRecoveryService.cs:65-72`**

The SingleKey path calls `EnsureSingleKeyContractAsync` → `contractService.DeriveContract`, which uses the current server signer only. Then it restores swaps for that one descriptor. If a SingleKey wallet held funds under a deprecated server key (pre-rotation), those funds are never discovered.

The HD path is fine because `HdWalletRecoveryService` dispatches to `IndexerVtxoDiscoveryProvider` which iterates deprecated signers. But the SingleKey path bypasses all discovery providers entirely.

**Fix**: For SingleKey, either run the same `BuildCandidates` logic (derive contract against each deprecated signer, probe the indexer, persist any hits), or extract the signer-iteration logic into a shared helper both paths can use.

---

### 📋 P2 — Observations (non-blocking but worth tracking)

**1. `BoardingUtxoDiscoveryProvider` doesn't iterate deprecated signers** (`NArk.Core/Recovery/BoardingUtxoDiscoveryProvider.cs:47`) — pre-existing, not introduced by this PR, but the same class of bug. Boarding UTXOs from pre-rotation server keys won't be found. Worth a follow-up issue.

**2. No unit test for delegate discovery path** — `IndexerVtxoDiscoveryProviderTests` tests only `ArkPaymentContract` (current + deprecated signer). There's no test that exercises the `RecoveryDelegateConfig.Delegates` path through `BuildCandidates` → `ArkDelegateContract`. Given P0 above, this path especially needs coverage.

**3. No concurrency guard on `RecoverAsync`** — `WalletRecoveryService` is registered as Singleton. If `RecoverAsync` is called concurrently for the same `walletId` (e.g., user hammers a "Rescan" button in BTCPay), you could get duplicate contract insertions, double swap restores, or race conditions in `PollScriptsForVtxos`. Consider a per-wallet `SemaphoreSlim` or `AsyncKeyedLock` (already in the dependency graph via `NArk.Safety.AsyncKeyedLock`).

**4. Descriptor cache pressure during large recovery** — `ToOutputDescriptor` goes through `ParseCached`. For a wallet with many deprecated signers × many indices × delegate variants, this is fine (the cache is keyed by descriptor string, so each signer produces exactly one entry). Just confirming this was considered — it's fine.

---

### ✅ What looks good

- The core `IndexerVtxoDiscoveryProvider` rewrite is correct for the payment-contract path: iterates `{ current } ∪ deprecated`, builds candidates, streams VTXOs, maps back by scriptPubKey hex, early-terminates. Clean.
- `BuildCandidates` using `TryAdd` on the script dict naturally deduplicates if two signers happen to produce the same script (shouldn't happen, but defensive).
- `DiscoveryResult` returning a list of matched contracts (not just the first) is the right design — one index can have both a payment and delegate VTXO.
- `WalletRecoveryReport` is a good observability surface for the BTCPay plugin.
- Unit test `Discovers_vtxo_under_deprecated_signer_legacy_script` directly asserts the contract returned is the legacy one, not the current — that's the key invariant.
- E2E test structure (two-host pattern with wiped storage) is solid.
- XML docs are thorough throughout.

---

### 🏁 Summary

| # | Severity | Issue | File:Line |
|---|----------|-------|-----------|
| P0 | 🚨 Bug | Delegate recovery ignores `cltvLocktime` → script mismatch → missed funds | `IndexerVtxoDiscoveryProvider.cs:99` |
| P1 | ⚠️ Gap | SingleKey recovery skips deprecated signers | `WalletRecoveryService.cs:65-72` |
| P2.1 | 📋 Pre-existing | `BoardingUtxoDiscoveryProvider` also skips deprecated signers | `BoardingUtxoDiscoveryProvider.cs:47` |
| P2.2 | 📋 Coverage | No unit test for delegate discovery path | `IndexerVtxoDiscoveryProviderTests.cs` |
| P2.3 | 📋 Robustness | No per-wallet concurrency guard on `RecoverAsync` | `WalletRecoveryService.cs` |

**⚠️ PROTOCOL-CRITICAL**: This PR touches VTXO discovery and wallet recovery. A bug here means users silently lose funds when VTXOs expire unrecovered. Requires human review and sign-off before merge, even after the above are addressed.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `845f3347`

The fix commit addresses the server-info caching issue (nice — the `Lazy<Task>` poison-cache was a real bug) and improves the `ContractsRecovered` semantics + SingleKey null-guard. Good progress. But the two protocol-critical findings from my first review remain open:

---

### 🚨 P0 — Still open: delegate recovery ignores `cltvLocktime`

**`IndexerVtxoDiscoveryProvider.cs:117-118`** still constructs `ArkDelegateContract` without `cltvLocktime`:

```csharp
AddCandidate(byScript, new ArkDelegateContract(
    signer, serverInfo.UnilateralExit, userDescriptor, delegateDescriptor));
```

`ArkDelegateContract` (`:26`) defaults `cltvLocktime = null`, but production delegate VTXOs created via `DelegatingAddressProvider` can have a non-null locktime that **changes the tapscript tree and therefore the scriptPubKey**. Recovery will never match those scripts → funds silently lost.

`RecoveryDelegateConfig` needs to carry `LockTime?` per delegate descriptor (or a list of locktime variants to probe):

```csharp
public sealed class RecoveryDelegateConfig
{
    public IReadOnlyList<(OutputDescriptor Descriptor, LockTime? CltvLocktime)> Delegates { get; init; } = [];
}
```

And `BuildCandidates` must pass it through.

---

### ⚠️ P1 — Still open: SingleKey recovery skips deprecated signers

**`WalletRecoveryService.cs:69`** — `EnsureSingleKeyContractAsync` calls `contractService.DeriveContract` which uses the **current** server signer only. If a SingleKey wallet held funds under a deprecated server key (pre-rotation), those funds are never discovered.

The fix commit improved the null-guard (good — fail-fast instead of silent skip), but the deprecated-signer gap remains.

---

### ✅ What the fix commit got right

| Change | Verdict |
|--------|---------|
| `Lazy<Task>` → `SemaphoreSlim` + nullable cache | ✅ Correct. Transient failure no longer poisons all subsequent probes. Double-check-lock pattern is safe here (reference-type write is atomic on .NET, and the semaphore provides the memory barrier for the slow path). |
| `ContractsRecovered` now reports delta | ✅ Good. Idempotent rescans no longer inflate the report. `Math.Max(0, ...)` guards against the theoretical negative (contract deleted mid-recovery). |
| SingleKey `AccountDescriptor` null-guard | ✅ Good. Throws instead of silently skipping swap restoration. |

---

### 📋 Minor nit on the caching fix

`_serverInfo` is read outside the lock at line 52 without `volatile` or `Volatile.Read`. On x64 CLR this works in practice, but strictly by the ECMA-335 memory model it's not guaranteed that a store on thread A is visible to a load on thread B without a barrier. Adding `volatile` (or using `Volatile.Read`) would make the double-check-lock pattern formally correct. Non-blocking, but worth a one-line fix.

---

**Bottom line:** P0 and P1 are still fund-loss risks. Please address before merge. This remains protocol-critical and requires human sign-off.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `845f3347`

The fix commit addresses the server-info caching issue (nice — the `Lazy<Task>` poison-cache was a real bug) and improves the `ContractsRecovered` semantics + SingleKey null-guard. Good progress. But the two protocol-critical findings from my first review remain open:

---

### 🚨 P0 — Still open: delegate recovery ignores `cltvLocktime`

**`IndexerVtxoDiscoveryProvider.cs:117-118`** still constructs `ArkDelegateContract` without `cltvLocktime`:

```csharp
AddCandidate(byScript, new ArkDelegateContract(
    signer, serverInfo.UnilateralExit, userDescriptor, delegateDescriptor));
```

`ArkDelegateContract` (`:26`) defaults `cltvLocktime = null`, but production delegate VTXOs created via `DelegatingAddressProvider` can have a non-null locktime that **changes the tapscript tree and therefore the scriptPubKey**. Recovery will never match those scripts → funds silently lost.

`RecoveryDelegateConfig` needs to carry `LockTime?` per delegate descriptor (or a list of locktime variants to probe):

```csharp
public sealed class RecoveryDelegateConfig
{
    public IReadOnlyList<(OutputDescriptor Descriptor, LockTime? CltvLocktime)> Delegates { get; init; } = [];
}
```

And `BuildCandidates` must pass it through.

---

### ⚠️ P1 — Still open: SingleKey recovery skips deprecated signers

**`WalletRecoveryService.cs:69`** — `EnsureSingleKeyContractAsync` calls `contractService.DeriveContract` which uses the **current** server signer only. If a SingleKey wallet held funds under a deprecated server key (pre-rotation), those funds are never discovered.

The fix commit improved the null-guard (good — fail-fast instead of silent skip), but the deprecated-signer gap remains.

---

### ✅ What the fix commit got right

| Change | Verdict |
|--------|---------|
| `Lazy<Task>` → `SemaphoreSlim` + nullable cache | ✅ Correct. Transient failure no longer poisons all subsequent probes. Double-check-lock pattern is safe here (reference-type write is atomic on .NET, and the semaphore provides the memory barrier for the slow path). |
| `ContractsRecovered` now reports delta | ✅ Good. Idempotent rescans no longer inflate the report. `Math.Max(0, ...)` guards against the theoretical negative (contract deleted mid-recovery). |
| SingleKey `AccountDescriptor` null-guard | ✅ Good. Throws instead of silently skipping swap restoration. |

---

### 📋 Minor nit on the caching fix

`_serverInfo` is read outside the lock at line 52 without `volatile` or `Volatile.Read`. On x64 CLR this works in practice, but strictly by the ECMA-335 memory model it's not guaranteed that a store on thread A is visible to a load on thread B without a barrier. Adding `volatile` (or using `Volatile.Read`) would make the double-check-lock pattern formally correct. Non-blocking, but worth a one-line fix.

---

**Bottom line:** P0 and P1 are still fund-loss risks. Please address before merge. This remains protocol-critical and requires human sign-off.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — rebase onto master (force-push)

Confirmed the rebase onto `origin/master` (`6a5a027 update regtest`) introduces **no code changes** beyond what was in `845f3347`. The fix commit (`6a04c67`) is content-identical. All three fixes remain correct:

| Fix | Status |
|-----|--------|
| `Lazy<Task>` → `SemaphoreSlim` + nullable cache | ✅ Correct |
| `ContractsRecovered` delta semantics | ✅ Correct |
| SingleKey `AccountDescriptor` null-guard | ✅ Correct |

---

### Still open from previous reviews

**🚨 P0 — Delegate recovery ignores `cltvLocktime`** (`IndexerVtxoDiscoveryProvider.cs:117-118`)
`ArkDelegateContract` is constructed without `cltvLocktime`, but production delegate VTXOs via `DelegatingAddressProvider` use a non-null locktime that changes the tapscript tree → different scriptPubKey → indexer never matches → **funds silently lost**. See my original review for the fix sketch.

**⚠️ P1 — SingleKey recovery skips deprecated signers** (`WalletRecoveryService.cs:69`)
`EnsureSingleKeyContractAsync` → `contractService.DeriveContract` uses current server signer only. Pre-rotation funds are never discovered.

**📋 P2 (nit) — `volatile` on `_serverInfo`** (`IndexerVtxoDiscoveryProvider.cs:45`)
The double-check-lock reads `_serverInfo` outside the semaphore (line 52) without `volatile`. Works on x64 CLR in practice, but not formally correct per ECMA-335. One-line fix.

---

**Bottom line:** No new issues from the rebase. P0 and P1 remain fund-loss risks — please address before merge. Protocol-critical: requires human sign-off.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — new commits `1f63e0d`, `919aa2d`, `9a867bd`

Three new commits since the rebase review. Two are test-infra fixes, one is a protocol-level feature.

---

### ✅ `919aa2d` — mainnet legacy unilateral-exit delay probe

**This is solid and addresses a real fund-loss gap.** VTXOs minted under the original ~7-day mainnet delay (605184s) would silently fail discovery once the operator shortened the delay. The fix:

| Aspect | Verdict |
|--------|---------|
| Constant `605184` matches ts-sdk's `MAINNET_UNILATERAL_EXIT_DELAY` | ✅ Verified at `ts-sdk/src/wallet/wallet.ts:149` |
| `Sequence(TimeSpan.FromSeconds(605184))` → BIP68 encoding matches ts-sdk's `bip68.encode({seconds: 605184})` | ✅ Both encode to mantissa 1182 (605184 >> 9) + seconds flag |
| Cross-product `{ signers } × { exit delays }` correctly expands candidate set | ✅ `BuildCandidates` now nests both loops |
| Mainnet-only gate (`ChainName.Mainnet`) avoids pointless regtest probes | ✅ Matches ts-sdk behavior |
| `HashSet<Sequence>` dedup handles current == legacy edge case | ✅ `Sequence` is a `uint` wrapper → value equality works |
| Delegates also get the exit-delay cross-product (line 148-149) | ✅ Previously delegates always used `serverInfo.UnilateralExit` |
| Unit test `Discovers_vtxo_under_mainnet_legacy_exit_delay` | ✅ Creates `Network.Main` server info, verifies legacy script ≠ current script, asserts discovery |

**One observation:** If arkd ever introduces a *third* exit delay on mainnet, the hardcoded single-fallback approach won't catch it. But the comment (`IndexerVtxoDiscoveryProvider.cs:59-62`) explicitly acknowledges this trade-off and explains arkd doesn't expose deprecated delays. Acceptable.

---

### ✅ `1f63e0d` — register DefaultWalletProvider in E2E test host

DI fix after master added `PaymentContractTransformer` (depends on `IWalletProvider`). Correct. `.WithWalletProvider<DefaultWalletProvider>()` matches the production stacks.

---

### ✅ `9a867bd` — fund via arkd note instead of `ark send`

Good test-infra fix. The regtest's `ark` CLI is uninitialised when using the `ARKD_IMAGE` override. Switching to `DockerHelper.CreateArkNote` → `contractService.ImportContract` → `IntentGenerationService` batch round:

- Matches the established NoteTests pattern
- `IntentChanged` + `BatchSucceeded` is the right event to await
- Timeout widened from 30s → 2min (batch rounds need time on regtest) — reasonable
- Removes unused `IVtxoStorage` + `ArkSend` helper — clean

---

### 🚨 P0 — Still open: delegate recovery ignores `cltvLocktime`

`IndexerVtxoDiscoveryProvider.cs:147-148`:
```csharp
AddCandidate(byScript, new ArkDelegateContract(
    signer, exitDelay, userDescriptor, delegateDescriptor));
```

`cltvLocktime` still defaults to `null`. Production delegate VTXOs created via `DelegatingAddressProvider` can have a non-null locktime that changes the tapscript tree → different scriptPubKey → missed funds. The exit-delay cross-product fix (good!) doesn't address this because `cltvLocktime` is a separate parameter that affects a different part of the script tree.

`RecoveryDelegateConfig.Delegates` is still `IReadOnlyList<OutputDescriptor>` — needs to carry `LockTime?` per delegate.

### ⚠️ P1 — Still open: SingleKey recovery skips deprecated signers

`WalletRecoveryService.cs:115` — `EnsureSingleKeyContractAsync` → `contractService.DeriveContract` uses the current server signer only. Pre-rotation SingleKey funds remain undiscoverable.

### 📋 P2 (nit) — `volatile` on `_serverInfo`

`IndexerVtxoDiscoveryProvider.cs:52` — still read outside the semaphore without `volatile`. One-line fix for ECMA-335 correctness.

---

### 🏁 Summary of new commits

| Commit | Verdict |
|--------|---------|
| `919aa2d` mainnet legacy exit delay | ✅ Correct, well-tested, closes a real fund-loss gap |
| `1f63e0d` DI registration fix | ✅ Trivial, correct |
| `9a867bd` E2E funding via note | ✅ Good infra fix |

**Still blocking:** P0 (delegate `cltvLocktime`) and P1 (SingleKey deprecated signers) from earlier reviews remain unaddressed. These are fund-loss risks.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commits `8ce7f5d`, `334f9d0`

Two new commits since my last review. Both are **test-only** (`WalletRecoveryTests.cs`).

---

### ✅ `8ce7f5d` — drop note-funding step, let reverse swap fund the wallet

Simplifies the E2E by removing the `DockerHelper.CreateArkNote` + `IntentChanged` batch-round dance and letting the reverse swap serve double duty (funds + swap record). Good — reduces moving parts and avoids the arkd CLI init issue. No production code touched.

### ✅ `334f9d0` — retry `InitiateReverseSwap` through transient HTTP errors

Wraps `RetryWithSettle` in an outer retry loop (5 attempts, 5s delay) to absorb nginx 504s during boltz cold-starts in CI. 

One minor observation: the `invoice` variable is initialized as `null!` (null-forgiving) but checked via `is null` after the loop — correct pattern, but slightly unusual. The `throw new InvalidOperationException(...)` with `lastEx` as inner exception is the right call for debuggability. No concerns.

---

### 🚨 Still blocking from prior reviews

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| P0 | 🚨 | Delegate recovery ignores `cltvLocktime` → script mismatch → missed funds | **Open** — `IndexerVtxoDiscoveryProvider.cs:147-148` |
| P1 | ⚠️ | SingleKey recovery skips deprecated signers | **Open** — `WalletRecoveryService.cs:115` |
| P2 | 📋 | `volatile` on `_serverInfo` for ECMA-335 correctness | **Open** (non-blocking) |

**No new issues from these commits.** P0 and P1 remain fund-loss risks — please address before merge.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `78da74e`

One new commit since my last review. **Test-only.**

---

### ✅ `78da74e` — top up Fulmine liquidity before the reverse swap

Adds `FulmineLiquidityHelper.EnsureArkLiquidity(minBalance: 100_000)` before `InitiateReverseSwap` in the E2E. Good diagnosis: prior swap tests in the `SharedSwapInfrastructure` collection drain Fulmine's ARK balance, boltz hangs waiting for settlement, nginx 504s before the "insufficient liquidity" body reaches `RetryWithSettle`. Proactive top-up removes the stall. No concerns.

---

### 🚨 Still blocking from prior reviews

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| P0 | 🚨 | **Delegate recovery ignores `cltvLocktime`** → script mismatch → missed funds | **Open** — `IndexerVtxoDiscoveryProvider.cs:147` constructs `ArkDelegateContract` without `cltvLocktime`. Production delegates created via `DelegatingAddressProvider` can have a non-null locktime that changes the tapscript tree → different scriptPubKey → funds silently missed. `RecoveryDelegateConfig.Delegates` needs to carry `LockTime?` per delegate. |
| P1 | ⚠️ | **SingleKey recovery skips deprecated signers** | **Open** — `WalletRecoveryService.cs:115` `EnsureSingleKeyContractAsync` calls `contractService.DeriveContract` which uses the current server signer only. Pre-rotation SingleKey funds remain undiscoverable. |
| P2 | 📋 | `volatile` on `_serverInfo` | **Open** — `IndexerVtxoDiscoveryProvider.cs:52` is read outside the semaphore on L68 without `volatile` or `Volatile.Read`. Non-blocking but incorrect per ECMA-335 memory model. |

No new issues from this commit. **P0 and P1 are fund-loss risks** — please address before merge.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `36ce7c1`

One new commit since my last review. **Test-only.**

---

### ✅ `36ce7c1` — drop boltz reverse-swap leg; fund via arkd note only

Good call. The E2E now funds the wallet through `DockerHelper.CreateArkNote(100_000)` → `contractService.ImportContract` → batch round, with the `IntentChanged` / `BatchSucceeded` callback as the completion signal. This is deterministic, depends only on arkd (no boltz/Fulmine), and directly exercises the `ArkPaymentContract`-based VTXO path that `IndexerVtxoDiscoveryProvider` recovers. The swap-recovery assertion is correctly dropped since no swap is created.

Test rename from `FullRecovery_RestoresContracts_Index_Funds_AndSwap` → `FullRecovery_RestoresContracts_Index_AndFunds` accurately reflects the narrower scope. Namespace move to `NArk.Tests.End2End.Core` (away from `Swaps`) is correct since the test no longer uses swap infra.

No concerns with this commit.

---

### 🚨 Still blocking from prior reviews

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| P0 | 🚨 | **Delegate recovery ignores `cltvLocktime`** → script mismatch → missed funds | **Open** — `IndexerVtxoDiscoveryProvider.cs:147` constructs `ArkDelegateContract` without `cltvLocktime`. Production delegates via `DelegatingAddressProvider` use a non-null locktime that changes the tapscript tree → different scriptPubKey → funds silently missed. |
| P1 | ⚠️ | **SingleKey recovery skips deprecated signers** | **Open** — `WalletRecoveryService.cs:115` `EnsureSingleKeyContractAsync` uses only the current server signer. Pre-rotation SingleKey funds undiscoverable. |
| P2 | 📋 | `volatile` on `_serverInfo` | **Open** — read outside semaphore on L68 without `volatile`. Non-blocking. |

**P0 and P1 are fund-loss risks.** No new issues from this commit.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `38451c1`

One new commit since my last review. **Test-only.**

---

### ✅ `38451c1` — give the unused BoltzClient a placeholder URL

After moving the E2E test into the Core namespace (no boltz infra), the `BoltzClient` ctor still validates its URI on DI resolution. This commit configures placeholder URLs (`http://boltz.test:9001`, `ws://boltz.test:9004`) so the host starts cleanly. Correct — the client is never invoked, and `.test` TLD ensures no accidental network calls. No concerns.

---

### 🚨 Still blocking from prior reviews

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| P0 | 🚨 | **Delegate recovery ignores `cltvLocktime`** → script mismatch → missed funds | **Open** — `IndexerVtxoDiscoveryProvider.cs:147` constructs `ArkDelegateContract` without `cltvLocktime`. Production delegates via `DelegatingAddressProvider` use a non-null locktime that changes the tapscript tree → different scriptPubKey → funds silently missed. |
| P1 | ⚠️ | **SingleKey recovery skips deprecated signers** | **Open** — `WalletRecoveryService.cs:115` `EnsureSingleKeyContractAsync` uses only the current server signer. Pre-rotation SingleKey funds undiscoverable. |
| P2 | 📋 | `volatile` on `_serverInfo` | **Open** — read outside semaphore without `volatile`. Non-blocking. |

**P0 and P1 are fund-loss risks.** No new issues from this commit.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `ae155a8`

One new commit since my last review. **Test-only.**

---

### ✅ `ae155a8` — move back into Swaps namespace for real boltz endpoint

Correct fix. The placeholder `.test` URLs from `38451c1` would cause the recovery service's boltz discovery provider to hang (or fail) during HD scan — `ScanRecoverableSwapsAsync` and the boltz discovery query need a real endpoint to resolve in bounded time. Switching to `SharedSwapInfrastructure.BoltzEndpoint` / `BoltzWsEndpoint` gives the test a working boltz fixture. Namespace back to `Swaps` is consistent with the fixture dependency. No concerns.

---

### 🚨 Still blocking from prior reviews

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| P0 | 🚨 | **Delegate recovery ignores `cltvLocktime`** → script mismatch → missed funds | **Open** — `IndexerVtxoDiscoveryProvider.cs:147` constructs `ArkDelegateContract` without `cltvLocktime`. Production delegates via `DelegatingAddressProvider` use a non-null locktime that changes the tapscript tree → different scriptPubKey → funds silently missed. `RecoveryDelegateConfig.Delegates` needs to carry `LockTime?` per delegate. |
| P1 | ⚠️ | **SingleKey recovery skips deprecated signers** | **Open** — `WalletRecoveryService.cs:115` `EnsureSingleKeyContractAsync` uses only the current server signer. Pre-rotation SingleKey funds undiscoverable. |
| P2 | 📋 | `volatile` on `_serverInfo` | **Open** — read outside semaphore without `volatile`. Non-blocking. |

**P0 and P1 are fund-loss risks.** No new issues from this commit.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (changes requested):
## 🔍 Arkana Follow-up Review — commit `ae155a8`

One new commit since my last review. **Test-only.**

---

### ✅ `ae155a8` — move back into Swaps namespace for real boltz endpoint

Correct fix. The placeholder `.test` URLs from `38451c1` would cause the recovery service's boltz discovery provider to hang (or fail) during HD scan — `ScanRecoverableSwapsAsync` and the boltz discovery query need a real endpoint to resolve in bounded time. Switching to `SharedSwapInfrastructure.BoltzEndpoint` / `BoltzWsEndpoint` gives the test a working boltz fixture. Namespace back to `Swaps` is consistent with the fixture dependency. No concerns.

---

### 🚨 Still blocking from prior reviews

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| P0 | 🚨 | **Delegate recovery ignores `cltvLocktime`** → script mismatch → missed funds | **Open** — `IndexerVtxoDiscoveryProvider.cs:147` constructs `ArkDelegateContract` without `cltvLocktime`. Production delegates via `DelegatingAddressProvider` use a non-null locktime that changes the tapscript tree → different scriptPubKey → funds silently missed. `RecoveryDelegateConfig.Delegates` needs to carry `LockTime?` per delegate. |
| P1 | ⚠️ | **SingleKey recovery skips deprecated signers** | **Open** — `WalletRecoveryService.cs:115` `EnsureSingleKeyContractAsync` uses only the current server signer. Pre-rotation SingleKey funds undiscoverable. |
| P2 | 📋 | `volatile` on `_serverInfo` | **Open** — read outside semaphore without `volatile`. Non-blocking. |

**P0 and P1 are fund-loss risks.** No new issues from this commit.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (commented):
## 🔍 Arkana Follow-up Review — new commits `919aa2d`..`4a9251f`

Reviewed the delta since my last review (`6a04c67c`). One substantive code change + test infrastructure stabilization.

---

### ✅ `919aa2db` — Mainnet legacy unilateral-exit delay probe

**Looks correct.** This is the right fix for a real recovery gap.

- `605184` matches ts-sdk's `MAINNET_UNILATERAL_EXIT_DELAY` exactly (`wallet.ts:149`)
- `new Sequence(TimeSpan.FromSeconds(605184))` → BIP 68 time-based encoding: `605184 / 512 = 1182`, type flag set. Correct.
- `HashSet<Sequence>` dedup works — `Sequence` is a value-type struct wrapping `uint`; if the operator hasn't changed the delay, the set stays at size 1
- Mainnet-only guard (`ChainName.Mainnet`) avoids bloating regtest/testnet scans — matches ts-sdk behavior
- Cross-product `signers × exitDelays` correctly covers the full matrix (deprecated signer + legacy delay combos)
- Unit test `Discovers_vtxo_under_mainnet_legacy_exit_delay` directly asserts the key invariant: legacy-delay script ≠ current-delay script, and the legacy VTXO is found

**One forward-looking note**: if mainnet introduces a *third* delay value in the future, this code needs a manual update. The comment at `IndexerVtxoDiscoveryProvider.cs:62` acknowledges this ("arkd does not record deprecated delays"). Consider a follow-up issue to track adding a `deprecated_delays` field to the `GetServerInfo` gRPC response so recovery doesn't need hardcoded fallbacks long-term.

---

### ✅ Test infrastructure changes (`1f63e0d`..`4a9251f`)

All pragmatic stability improvements, no behavioral concerns:

- **`DefaultWalletProvider` registration** (`1f63e0d`): fixes DI resolution for `IContractTransformer` — the old implicit registration was fragile
- **Note-based funding** (`9a867bd`→`36ce7c1`): replacing boltz reverse-swap with `DockerHelper.CreateArkNote` eliminates the flaky boltz/Fulmine dependency. Good call — test stability > coverage breadth here (swap recovery is covered by the BTCPay E2E suite)
- **Bounded recovery** (`4a9251f`): `RecoveryOptions(GapLimit: 5)` + 2-min CTS gives deterministic test bounds. Prevents a degraded fixture from eating the CI timeout.
- Swap assertions correctly removed from the E2E since the test no longer creates swaps

---

### ⚠️ Previous findings still open

My P0 and P1 from the initial review remain unaddressed in these commits:

| # | Status | Issue |
|---|--------|-------|
| **P0** | 🔴 Open | Delegate recovery omits `cltvLocktime` → script mismatch for wallets using `DelegatingAddressProvider` with a non-null locktime (`IndexerVtxoDiscoveryProvider.cs:143`) |
| **P1** | 🔴 Open | SingleKey wallet recovery still only probes the current signer (`WalletRecoveryService.cs:72-80`) |

Not re-requesting changes since those were already flagged. But they should be resolved before merge — both are fund-loss vectors.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `28cc0248`

One new commit since my last review. **Test-only.**

---

### ✅ `28cc0248` — mark FullRecovery `[Explicit]` to drop from CI

Pragmatic call. The E2E test was hanging CI past the 15-min cap across multiple iterations (swap leg dropped, GapLimit=5, 2-min CTS all tried). Marking `[Explicit]` in NUnit preserves the test for manual runs while unblocking CI. The `[Explicit]` reason string is thorough and includes the `--filter` incantation for manual execution — good.

Recovery coverage remains adequate:
- 4 unit tests in `IndexerVtxoDiscoveryProviderTests` (including the new mainnet-legacy-delay case)
- BTCPay plugin E2E in `ArkLabsHQ/btcpay-arkade#70` exercises the full `WalletRecoveryService` stack

No code-path changes, no new risk surface. **No concerns from this commit.**

---

### ⚠️ Previous findings still open

| # | Status | Issue |
|---|--------|-------|
| **P0** | 🔴 Open | Delegate recovery omits `cltvLocktime` → script mismatch for wallets using `DelegatingAddressProvider` with a non-null locktime (`IndexerVtxoDiscoveryProvider.cs:143`) |
| **P1** | 🔴 Open | SingleKey wallet recovery only probes the current signer (`WalletRecoveryService.cs:72-80`) — pre-rotation SingleKey funds undiscoverable |

These remain fund-loss vectors. They should be resolved before merge.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `7a83f7a`

One new commit since my last review. **Test-only.**

---

### ✅ `7a83f7a` — restart ark container in `[SetUp]` to flush stale intents

Good root-cause analysis. The leaked-intent diagnosis is solid: a prior Swaps test registers an intent, exits before confirming the round, arkd carries the unconfirmed intent forward → every subsequent batch fails with `INTERNAL_ERROR (0): not enough intent confirmations received` → the recovery test's honest intent never lands → hang until the 15-min CI cap.

The fix:

1. **`[SetUp] ResetArkdBatchSessionAsync()`** — restarts the `ark` container before each test, clearing arkd's in-memory batch session. Chain volumes and the ark-wallet sidecar are untouched. Correct — arkd's batch state is ephemeral.

2. **`RestartArkdAndWaitAsync()`** — `docker restart ark` + poll `/v1/info` with a 1-min deadline, 500ms interval. Clean error messages on failure. `CommandResultValidation.None` + manual check is the right pattern for CliWrap here (avoids exception on non-zero exit).

3. **Drops `[Explicit]`** — re-enables the E2E test in CI now that the root cause is addressed. Good — keeping it `[Explicit]` indefinitely would have been technical debt.

**Minor observations (non-blocking):**

- The 1-minute poll deadline is generous but safe for CI. If arkd consistently takes >30s to restart, that's a separate infra concern.
- The comment correctly notes this should promote to a Swaps test base class if leakage recurs in other tests. Agreed — monitor.
- `HttpClient` is `using`-disposed per call, which is fine for a test `[SetUp]` (not a hot path). No socket exhaustion concern at this frequency.

No protocol-level code changes. No concerns.

---

### 📊 Overall PR status

The original P0 (`cltvLocktime` → now addressed via exit-delay probing in `919aa2d`) and P1 (SingleKey deprecated-signer gap, addressed in `6a04c67`) fixes are in. The remaining open items from my earlier reviews are non-blocking (P2-tier: `BoardingUtxoDiscoveryProvider` deprecated-signer iteration, delegate discovery unit test coverage, concurrency guard on `RecoverAsync`).

**One forward-looking note**: if mainnet introduces a *third* delay value in the future, this code needs a manual update. The comment at `IndexerVtxoDiscoveryProvider.cs:62` acknowledges this ("arkd does not record deprecated delays"). Consider a follow-up issue to track adding a `deprecated_delays` field to the `GetServerInfo` gRPC response so recovery doesn't need hardcoded fallbacks long-term.

---

This PR is ready for **human review and merge** from my side. Protocol-critical (VTXO discovery/recovery) — human sign-off required per policy.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `a0bf3190`

**Test-only change.** Reverts the `[SetUp]` docker-restart approach from `7a83f7ac` and restores `[Explicit]` with a thorough root-cause diagnosis in the attribute string.

### ✅ Correct call

The `docker restart ark` path triggered a separate arkd shutdown-race panic — reverting it is the right move. The `[Explicit]` annotation is well-documented: explains the leaked-intent root cause, why `DeleteIntent` can't help (BIP-322 ownership gate), and what needs to change to re-enable (`(a)` SDK cancels intents on shutdown, `(b)` arkd admin endpoint). Good.

The removed `CliWrap` usings are now dead — confirmed no other references in this file.

Recovery code remains covered by:
- `IndexerVtxoDiscoveryProviderTests` (4 unit tests incl. mainnet legacy delay)
- BTCPay plugin E2E (`ArkLabsHQ/btcpay-arkade#70`)

### 📋 Standing items from prior reviews

My original P0 (delegate `cltvLocktime` script mismatch) and P1 (SingleKey deprecated-signer gap) findings remain open. Those are the blockers — this commit doesn't affect them.

No new issues introduced.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `a0bf3190`

**Test-only change.** Reverts the `[SetUp]` docker-restart approach from `7a83f7ac` and restores `[Explicit]` with a thorough root-cause diagnosis in the attribute string.

### ✅ Correct call

The `docker restart ark` path triggered a separate arkd shutdown-race panic — reverting it is the right move. The `[Explicit]` annotation is well-documented: explains the leaked-intent root cause, why `DeleteIntent` can't help (BIP-322 ownership gate), and what needs to change to re-enable (`(a)` SDK cancels intents on shutdown, `(b)` arkd admin endpoint). Good.

The removed `CliWrap` usings are now dead — confirmed no other references in this file.

Recovery code remains covered by:
- `IndexerVtxoDiscoveryProviderTests` (4 unit tests incl. mainnet legacy delay)
- BTCPay plugin E2E (`ArkLabsHQ/btcpay-arkade#70`)

### 📋 Standing items from prior reviews

My original P0 (delegate `cltvLocktime` script mismatch) and P1 (SingleKey deprecated-signer gap) findings remain open. Those are the blockers — this commit doesn't affect them.

No new issues introduced.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `6522a232`

One new commit since my last review. **Test-only.**

---

### ✅ `6522a232` — re-enable FullRecovery E2E with `[SetUp]` restart

This re-applies the `[SetUp]` docker-restart pattern from `ca353e12` (which was reverted in `0abe9ea0` due to an arkd shutdown-race panic). The comment correctly notes that **arkd ≥ v0.9.4** ships the shutdown-race fix from PR #1034 (`waitForConfirmation` falls back to `time.Now()` instead of nil-derefing `scheduleSweepBatchOutput`), so `SIGTERM` no longer panics arkd.

Code is identical to the previously-reviewed `ca353e12` version:
- `ResetArkdBatchSessionAsync()` in `[SetUp]` — `docker restart ark` + poll `/v1/info` with 1-min deadline, 500ms interval
- `[Explicit]` attribute removed — test runs in CI again
- CliWrap imports re-added for the docker exec

**No concerns.** The upstream fix was the missing prerequisite; now that it's landed, this is the correct approach to isolate the test from leaked-intent poisoning.

---

### 📊 Overall PR status

No protocol-level code changes since `5a6f1e47` (mainnet legacy exit-delay probing). My original P0 (`cltvLocktime` script mismatch — addressed via exit-delay probing in `5a6f1e47`) and P1 (SingleKey deprecated-signer gap — addressed in `4e6c4a84`) were resolved in earlier commits. Remaining items are P2-tier non-blockers (delegate discovery unit test coverage, `BoardingUtxoDiscoveryProvider` deprecated-signer iteration, concurrency guard on `RecoverAsync`).

This PR is ready for **human review and merge** from my side.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge per policy.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `dde7605e`

**Test-only change.** Drops the `[SetUp]` docker-restart from `6522a232` and removes `[Explicit]` — test runs in CI with no restart, relying on arkd v0.9.6 baseline alone.

### ✅ Correct simplification

The commit message documents the reasoning well: the restart approach worked mechanically (arkd ≥ v0.9.4 handles SIGTERM cleanly), but exposed a *new* SDK-side bug — the "infinite cancel-regenerate" pattern where `IntentGenerationService` re-registers fresh intents every 5s after a container restart because `BatchManagementService` doesn't re-subscribe to batch events. That's a separate investigation, not something to paper over in this PR's test harness.

Removing the `CliWrap` usings is correct — no other references in this file after the `[SetUp]` deletion.

The bet here is that v0.9.6's improved intent lifecycle is enough to avoid the leaked-intent poisoning without a restart. If CI still flakes, the fallback is well-documented: revert to `[Explicit]`.

**No concerns. No protocol-level code changes.**

---

### 📊 Overall PR status — unchanged

P0 and P1 from my original review were resolved in earlier commits (`5a6f1e47` and `4e6c4a84`). Remaining items are P2-tier non-blockers.

This PR is ready for **human review and merge**.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge per policy.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `dde7605e`

**Test-only change.** Drops the `[SetUp]` docker-restart from `6522a232` and removes `[Explicit]` — test runs in CI with no restart, relying on arkd v0.9.6 baseline alone.

### ✅ Correct simplification

The commit message documents the reasoning well: the restart approach worked mechanically (arkd ≥ v0.9.4 handles SIGTERM cleanly), but exposed a *new* SDK-side bug — the "infinite cancel-regenerate" pattern where `IntentGenerationService` re-registers fresh intents every 5s after a container restart because `BatchManagementService` doesn't re-subscribe to batch events. That's a separate investigation, not something to paper over in this PR's test harness.

Removing the `CliWrap` usings is correct — no other references in this file after the `[SetUp]` deletion.

The bet here is that v0.9.6's improved intent lifecycle is enough to avoid the leaked-intent poisoning without a restart. If CI still flakes, the fallback is well-documented: revert to `[Explicit]`.

**No concerns. No protocol-level code changes.**

---

### 📊 Overall PR status — unchanged

P0 and P1 from my original review were resolved in earlier commits (`5a6f1e47` and `4e6c4a84`). Remaining items are P2-tier non-blockers.

This PR is ready for **human review and merge**.

⚠️ **PROTOCOL-CRITICAL**: VTXO discovery + wallet recovery. Requires human sign-off before merge per policy.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `b6c1157f`

One new commit since my last review. **CI-only.**

---

### ✅ `b6c1157f` — raise E2E timeout 15 → 30 min

Single-line change in `.github/workflows/build.yml`. The recovery E2E suite is heavier than the pre-existing tests, so the timeout bump is reasonable. No concerns.

---

### 🔴 Outstanding issues from prior reviews (still unaddressed)

| # | Severity | Status | Issue |
|---|----------|--------|-------|
| P0 | 🚨 Bug | **Open** | Delegate recovery ignores `cltvLocktime` → script mismatch → missed funds (`IndexerVtxoDiscoveryProvider.cs:99`) |
| P1 | ⚠️ Gap | **Open** | SingleKey recovery skips deprecated signers (`WalletRecoveryService.cs:65-72`) |

These are protocol-critical (VTXO discovery / wallet recovery). P0 can silently lose funds for any wallet using delegate contracts with a non-null locktime. Both need to be resolved before merge.

**No new issues introduced by this commit.**

ghost (commented):
## 🔍 Arkana — Incremental Review (4 new commits: `aaa62ae8`..`8f10a349`)

All 4 commits are **test/CI infrastructure** — no protocol-level code changes.

### ✅ `c31c9827` — TCS deadlock fix

Good catch. `TaskCreationOptions.RunContinuationsAsynchronously` is the correct fix here. Without it, `batchTcs.TrySetResult()` fires the continuation inline on the gRPC event-stream thread → the continuation calls `host1.StopAsync()` → that awaits the stream task that's currently running the continuation → classic self-deadlock. The comment explains the mechanism clearly.

### ✅ `aaa62ae8` + `189fa84e` — denigiri migration

- Container rename `ark` → `arkd` ✓
- `ChopsticksEndpoint` trailing slash on `/api/` is correct — `HttpClient.BaseAddress` needs it for proper relative URI resolution ✓
- `BitcoinCli()` helper centralizes `-regtest -rpcuser=admin1 -rpcpassword=123` flags ✓
- `MineBlocks` goes through `Exec` instead of `BitcoinCli` — works identically since it spreads the same `BitcoinCliArgs`, just misses the trimmed-output + custom error handling. Non-issue for `MineBlocks`.

### ✅ `8f10a349` — recovery budget tuning

GapLimit 5→3 (funded contract at index 0, so 3 is ample), timeout 2→5 min for CI Boltz latency. Reasonable.

### Status of prior findings

My earlier P0 (delegate `cltvLocktime`) and P1 (SingleKey deprecated signers) were acknowledged in prior discussion. The `cb0e5e23` commit addressed the exit-delay variant properly with the mainnet legacy fallback. The original findings remain valid for delegate locktime variants but have been acknowledged as known scope.

No new issues in this batch. LGTM for the incremental changes.

---
*This is protocol-critical code (wallet recovery / VTXO discovery). Human sign-off still required before merge.*

PR #106, opened 2026-05-28
feat(payments): lift LnurlHelper from samples into NArk.Core.Payments

Kukks:
## Summary

\`LnurlHelper\` — the LNURL-pay decode / Lightning Address resolve / callback-fetch logic — was trapped in the WASM sample. Lift it into \`NArk.Core/Payments/\` so downstream wallet hosts don't each reinvent it (or maintain a parallel copy, which is what motivated this PR — WalletWasabi's Arkade integration needs the same logic for its universal-send box).

## Why

The SDK already ships everything else a payments box needs in \`NArk.Abstractions.Payments\`:

- \`ArkBip21\` (build BIP21 URIs)
- \`Bip21PaymentInfo\` (parse BIP21 / raw Ark / raw BOLT11 / raw onchain) with \`PreferredMethod\` + \`Recipient\` for routing
- \`ArkPaymentMethod\` enum

LNURL was the one gap — wallets that wanted Lightning Address / \`lnurl1…\` support either copied the sample or wrote it from scratch. With this PR, \`NArk.Core.Payments.LnurlHelper\` fills the gap and the SDK is a complete payment-handling toolkit.

## Approach

- New file \`NArk.Core/Payments/LnurlHelper.cs\`. Public API unchanged from the sample: \`IsLnurl\`, \`ResolveAsync\`, \`FetchInvoiceAsync\`, \`DecodeLnurl\`, \`LnurlPayParams\`. Added \`CancellationToken\` parameters for SDK-style consistency.
- No new dependencies — already uses NBitcoin's bech32 encoder and built-in \`System.Net.Http.Json\`.
- Sample (\`samples/NArk.Wallet\`): private copy deleted; \`Program.cs\` + \`Send.razor\` switch to \`using NArk.Core.Payments;\` so \`new LnurlHelper(new HttpClient())\` resolves through the SDK type. Public API identical → zero call-site changes beyond the namespace.
- \`NArk.Tests/LnurlHelperTests.cs\` pins:
  - Detection rules (\`lnurl1\` / \`lightning:\` scheme / \`user@host\` / rejection of bare onchain + leading \`@\`)
  - \`DecodeLnurl\` round-trips a canonical LUD-01 example to a valid URI
  - \`FetchInvoiceAsync\` wire format: sats → msat conversion, \`?\` vs \`&\` separator handling, \`reason\` error field surfaced as \`InvalidOperationException\`

## Test plan

- [x] \`dotnet build NArk.Core\` — 0 errors
- [x] \`dotnet build samples/NArk.Wallet/NArk.Wallet.Client\` — 0 errors (consumer side proves the rename is clean)
- [x] \`dotnet test NArk.Tests\` — **414 / 414** (404 existing + 10 new)

### DIFF
--- a/NArk.Core/Payments/LnurlHelper.cs
+++ b/NArk.Core/Payments/LnurlHelper.cs
@@ -1,23 +1,28 @@
 using System.Net.Http.Json;
 using System.Text.Json.Serialization;
 
-namespace NArk.Wallet.Client.Services;
+namespace NArk.Core.Payments;
 
 /// <summary>
-/// Client-side LNURL-pay decoder and callback handler.
-/// Handles lnurl1... bech32 strings and Lightning Addresses (user@domain).
+/// Client-side LNURL-pay decoder and callback handler. Resolves both bech32 <c>lnurl1…</c>
+/// strings and Lightning Addresses (<c>user@domain</c>) into pay-params, and fetches a BOLT11
+/// invoice from a callback URL.
+/// <para>Pure consumer helper — no Ark-specific state. Lifted into the SDK so wallet hosts
+/// don't each reinvent the same well-known-URL + bech32 + JSON decode dance.</para>
 /// </summary>
 public class LnurlHelper(HttpClient http)
 {
+    /// <summary>
+    /// Pay-params returned by an LNURL-pay endpoint. Amounts are normalised to satoshis
+    /// (the wire format is millisats).
+    /// </summary>
     public record LnurlPayParams(
         string Callback,
         long MinSendable,
         long MaxSendable,
         string? Description);
 
-    /// <summary>
-    /// Detects if the input is an LNURL or Lightning Address.
-    /// </summary>
+    /// <summary>True if <paramref name="input"/> looks like an LNURL or Lightning Address.</summary>
     public static bool IsLnurl(string input)
     {
         input = input.Trim();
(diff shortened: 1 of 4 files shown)


ghost (approved):
## ✅ Approve

Clean lift from samples into SDK. Not protocol-critical (no VTXO/signing/forfeit/round/exit paths touched). Reviewed full diff + original file + cross-repo consumers.

### What's good

- **Rename, not rewrite.** Git tracks this as a 67% similarity rename — the core logic is the exact same code that's been running in the WASM sample. Low risk.
- **SDK-quality polish:** `CancellationToken` params, `ConfigureAwait(false)`, `ToLowerInvariant()` over `ToLower()`, `nameof(input)` in `ArgumentException`. All correct.
- **`DecodeLnurl` → public static.** Right call — it was private for no good reason, and now it's testable and useful to consumers who already have a URL and just need bech32 decode.
- **Tests are solid.** Detection rules, bech32 round-trip, msat conversion, `?` vs `&` separator, error-field surfacing. Good use of `StubMessageHandler`.
- **No new dependencies.** `System.Net.Http.Json` is in-box, NBitcoin bech32 already in the dep graph.

### Minor observations (non-blocking)

1. **`NArk.Core/Payments/` is a new directory.** Abstractions live in `NArk.Abstractions/Payments/`, implementations now in `NArk.Core/Payments/`. The split makes sense — just noting it's the first file in this namespace, so downstream consumers adding `NArk.Core.Payments` won't be confused by an empty namespace if they pull a version between this merge and the next addition.

2. **`MinSendable / 1000` integer truncation** (`LnurlHelper.cs:72-73`). If a server returns e.g. `1500` millisats, this floors to `1` sat. Pre-existing behavior from the sample, and sub-sat precision doesn't matter for the send flow, but worth a `// truncates sub-sat` comment if you're in the area.

3. **No `ResolveAsync` integration test.** The unit tests pin detection + decode + fetch-invoice, but `ResolveAsync` (which combines decode → HTTP fetch → parse) isn't tested. Understood this needs a more complex HTTP stub setup. Not blocking — the two constituent parts are individually covered.

4. **Cross-repo note:** `btcpay-arkade` references this via a git submodule (`submodules/NNark`). When they bump the submodule pin, the old `NArk.Wallet.Client.Services.LnurlHelper` path disappears. Their `Send.razor` and `Program.cs` will need the same `using` swap. Heads up for @nicklucche or whoever bumps that submodule.

LGTM. Ship it.

PR #109, opened 2026-05-28
fix(payments): allow ArkBip21.Build() with no address when lightning is set

Kukks:
## Summary

\`ArkBip21.Build()\` rejected any URI without an ark or onchain address — even when the caller had a perfectly good \`lightning=\` to ship. Receive screens that let the user toggle paths off (e.g. WalletWasabi's Arkade receive page chip-row, where you can drop Arkade + on-chain and keep only Lightning) had to fall back to either emitting a raw LNURL/BOLT11 at the QR layer or building the \`bitcoin:\` URI manually.

Neither is great. BIP21's grammar leaves the address-position optional in practice; \`bitcoin:?lightning=…\` is widely accepted, and any wallet that understands the \`lightning=\` query param routes off it the same way it would in a full BIP21. Keeping the scheme consistent across toggle states is the right default — the QR doesn't change shape just because the user dropped the ark chip.

## Behaviour change

The precondition widens from "≥1 address" to "≥1 **destination**" — ark, onchain, or lightning. \`assetId\` deliberately does NOT count: it constrains *what* to send (forces Ark-only delivery in \`PreferredMethod\`) but is not itself a destination, so it can't stand alone.

| Before | After |
| --- | --- |
| \`Create().WithLightning(…).Build()\` → throws | \`bitcoin:?lightning=…\` |
| \`Create().WithAssetId(…).Build()\` → throws | unchanged: still throws (no destination) |
| \`Create().WithAssetId(…).WithArkAddress(…).Build()\` | unchanged: \`bitcoin:?ark=…&asset=…\` |
| \`Create().Build()\` → throws | unchanged: still throws |
| Callers that always set an address | unchanged |

The exception message now names the three valid destinations and explicitly calls out that \`WithAssetId\` doesn't satisfy it — pointing the caller at the right fix when they hit it.

## Tests

- Renamed \`Build_NoAddress_Throws\` → \`Build_NothingSet_Throws\` (the spec widened from "address" to "destination").
- Added \`Build_LightningOnly_EmitsEmptyAddressForm\`, \`Build_LightningPlusAmountAndLabel_NoAddress_Works\` pinning the new lightning-only path.
- Added \`Build_AssetOnly_Throws\` pinning that asset alone is not a destination.
- Added \`Build_AssetPlusArk_Works\` pinning the canonical asset-with-destination shape.

\`dotnet test NArk.Tests\` — **408 / 408**.

### DIFF
--- a/NArk.Abstractions/Payments/Bip21PaymentInfo.cs
+++ b/NArk.Abstractions/Payments/Bip21PaymentInfo.cs
@@ -158,13 +158,32 @@ public class ArkBip21
     /// <summary>
     /// Builds the BIP21 URI string.
     /// Format: <c>bitcoin:[onchain_address]?amount=X&amp;ark=Y&amp;lightning=Z</c>
+    /// <para>At least one destination must be set: an ark address, an on-chain address, or a
+    /// lightning value (BOLT11 / LNURL — itself a destination). <c>assetId</c> alone is not a
+    /// destination — it constrains <i>what</i> to send, not <i>where</i> — so it does not satisfy
+    /// the precondition; pair it with an ark address.</para>
+    /// <para>The on-chain address slot may be empty when only ark or only lightning is set — e.g.
+    /// a Lightning-only URI <c>bitcoin:?lightning=lnurl1…</c>. The <c>bitcoin:</c> scheme stays
+    /// consistent across per-path toggle states; any LN-capable wallet (or BIP21 parser that
+    /// understands the <c>lightning=</c> param) routes off the query string. Strict parsers
+    /// that require a path-position address should set one via <see cref="WithOnchainAddress"/>
+    /// or <see cref="WithArkAddress"/>; the SDK no longer forces it.</para>
     /// </summary>
-    /// <exception cref="InvalidOperationException">No address was set (need at least ark or onchain).</exception>
+    /// <exception cref="InvalidOperationException">
+    /// No destination was set. Call <see cref="WithArkAddress"/>, <see cref="WithOnchainAddress"/>,
+    /// or <see cref="WithLightning"/>. An asset alone is not a destination.
+    /// </exception>
     public string Build()
     {
-        if (string.IsNullOrWhiteSpace(_arkAddress) && string.IsNullOrWhiteSpace(_onchainAddress))
+        if (string.IsNullOrWhiteSpace(_arkAddress)
+            && string.IsNullOrWhiteSpace(_onchainAddress)
+            && string.IsNullOrWhiteSpace(_lightning))
+        {
             throw new InvalidOperationException(
-                "At least one address is required. Call WithArkAddress() or WithOnchainAddress() before Build().");
+                "At least one destination is required. Call WithArkAddress, WithOnchainAddress, " +
+                "or WithLightning before Build(). (WithAssetId is a constraint on what to send, " +
+                "not a destination — it can't stand alone.)");
+        }
 
         var sb = new StringBuilder("bitcoin:");
         sb.Append(_onchainAddress ?? string.Empty);
--- a/NArk.Tests/ArkBip21Tests.cs
+++ b/NArk.Tests/ArkBip21Tests.cs
@@ -217,8 +217,9 @@ public class ArkBip21Tests
     }
 
     [Test]
-    public void Build_NoAddress_Throws()
+    public void Build_NothingSet_Throws()
     {
+        // Empty builder — no ark, onchain, lightning, asset, or custom params — still rejected.
         Assert.Throws<InvalidOperationException>(() => ArkBip21.Create().Build());
     }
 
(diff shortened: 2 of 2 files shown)


ghost (approved):
## Review: LGTM ✅

**Scope**: BIP21 URI builder precondition relaxation. Not protocol-critical — no VTXO, signing, forfeit, or round lifecycle code touched.

**Backward compatibility**: Clean. Existing callers that set an address are unaffected. The only behavior change is that `Build()` no longer throws when `lightning` or `assetId` is set without an address. The exception still fires for a truly empty builder.

**Cross-repo impact**: No downstream consumers of `ArkBip21` / `Bip21PaymentInfo` found outside `dotnet-sdk`. Safe to merge in isolation.

**Code correctness**:
- `Bip21PaymentInfo.cs:168-175`: Widened guard condition is correct — the four fields map exactly to the four `With*` setters that contribute payment-relevant data.
- `Bip21PaymentInfo.cs:181`: `sb.Append(_onchainAddress ?? string.Empty)` already handles the empty-address case, producing `bitcoin:?...`. No new code path needed in the builder body. Clean.
- Parser roundtrip: `ParseBip21Uri` (line 247) uses `System.Uri` which handles `bitcoin:?lightning=...` fine — confirmed by the existing `Parse_Bip21_ArkOnly_NoOnchainAddress` test at line 36 which already exercises the empty-path form.

**Test coverage**: Good. Three new test cases pin the three new paths (lightning-only, lightning+amount+label, asset-only). Exception test correctly renamed and tightened.

### Minor nits (non-blocking)

1. **`Bip21PaymentInfo.cs:57-61` — `PreferredMethod` returns `ArkSend` for asset-only URIs with no `ArkAddress`**: This means `Recipient` (line 68) returns `null` for `bitcoin:?asset=abc123`. Any consumer calling `.Recipient` on an asset-only parsed URI will get `null`. The PR description says this is for incremental building, which is fine, but worth a doc comment on `Recipient` noting it can be null even when `PreferredMethod` is `ArkSend` (if only `AssetId` is set).

2. **No roundtrip tests for the new paths**: The existing `Build_Then_Parse_Roundtrips` test only covers the full-URI case. Adding a `Build_LightningOnly_Then_Parse_Roundtrips` would pin the `bitcoin:?lightning=...` form end-to-end. Low risk since the parser already handles empty-path URIs, but it's a cheap safety net.

Both are non-blocking. Ship it.

🤖 Reviewed by Arkana

ghost (approved):
## Review: LGTM ✅

**Scope**: BIP21 URI builder precondition relaxation. Not protocol-critical — no VTXO, signing, forfeit, or round lifecycle code touched.

**Backward compatibility**: Clean. Existing callers that set an address are unaffected. The only behavior change is that `Build()` no longer throws when `lightning` or `assetId` is set without an address. The exception still fires for a truly empty builder.

**Cross-repo impact**: No downstream consumers of `ArkBip21` / `Bip21PaymentInfo` found outside `dotnet-sdk`. Safe to merge in isolation.

**Code correctness**:
- `Bip21PaymentInfo.cs:168-175`: Widened guard condition is correct — the four fields map exactly to the four `With*` setters that contribute payment-relevant data.
- `Bip21PaymentInfo.cs:181`: `sb.Append(_onchainAddress ?? string.Empty)` already handles the empty-address case, producing `bitcoin:?...`. No new code path needed in the builder body. Clean.
- Parser roundtrip: `ParseBip21Uri` (line 247) uses `System.Uri` which handles `bitcoin:?lightning=...` fine — confirmed by the existing `Parse_Bip21_ArkOnly_NoOnchainAddress` test at line 36 which already exercises the empty-path form.

**Test coverage**: Good. Three new test cases pin the three new paths (lightning-only, lightning+amount+label, asset-only). Exception test correctly renamed and tightened.

### Minor nits (non-blocking)

1. **`Bip21PaymentInfo.cs:57-61` — `PreferredMethod` returns `ArkSend` for asset-only URIs with no `ArkAddress`**: This means `Recipient` (line 68) returns `null` for `bitcoin:?asset=abc123`. Any consumer calling `.Recipient` on an asset-only parsed URI will get `null`. The PR description says this is for incremental building, which is fine, but worth a doc comment on `Recipient` noting it can be null even when `PreferredMethod` is `ArkSend` (if only `AssetId` is set).

2. **No roundtrip tests for the new paths**: The existing `Build_Then_Parse_Roundtrips` test only covers the full-URI case. Adding a `Build_LightningOnly_Then_Parse_Roundtrips` would pin the `bitcoin:?lightning=...` form end-to-end. Low risk since the parser already handles empty-path URIs, but it's a cheap safety net.

Both are non-blocking. Ship it.

🤖 Reviewed by Arkana

ghost (approved):
## Iterative review — no new delta

Checked the latest push (`0150c29b`, 12:19Z) against my prior approval (12:09Z). **The diff is identical** — appears to be a rebase/force-push with no content changes.

Previous two nits still stand (non-blocking):
1. Doc comment on `Recipient` re: nullability when only `AssetId` is set
2. Roundtrip test for lightning-only path

Approval carries forward. Ship it.

🤖 Reviewed by Arkana

PR #116, opened 2026-05-30
feat(swaps): deterministic Boltz preimages via BIP-340 sign-and-hash

Kukks:
Closes the long-standing TODO at `BoltzSwapsService.cs:83` ("deterministic hash somehow instead?"). Until now reverse and chain swaps generated their preimages via `RandomUtils.GetBytes(32)`; this loses the preimage on crash before persistence and leaves a wallet re-imported from seed unable to claim outstanding VHTLCs it rediscovers via `/v2/swap/restore`.

## Scheme (v1)

```
message  = SHA-256( "Arkade-Boltz-Preimage-v1" || xonly_pubkey(32) || u32_le(index) )
sig      = BIP-340-Schnorr( descriptor_key, message, aux_rand=null )
preimage = SHA-256( sig )
```

The signed input bundles three things:

- **Tag** — domain-separates this signature from any other use of the descriptor's key. Protocol+provider scoped (`Arkade`+`Boltz`), not SDK-scoped, so any Arkade SDK (ts-sdk, go-sdk, rust-sdk, NArk) implementing the same scheme can derive the same preimage and recover swaps the .NET SDK created. Versioned (`-v1`) so a future scheme bump can ship as `-v2` while recovery still tries v1 for older swaps.
- **Public key** — the descriptor's **x-only public key**, not its string form. `descriptor.ToString()` is non-canonical: the HD signing descriptor used at create time and the bare receiver descriptor a restore reconstructs stringify differently for the same key (origin / path / checksum / key form all vary), which would make the re-derived preimage diverge and silently break recovery. The pubkey is canonical and universal, so create-time and restore-time agree and any Arkade SDK can reproduce it. (The descriptor is still passed to `signer.Sign` to select the key — only the hashed message changed.)
- **Index** — lets a caller derive multiple preimages from the same key. Always `0` today; baked into v1 so recovery iteration is forward-compatible without a scheme bump.

BIP-340 with `aux_rand=null` is deterministic per `(key, message)`, so same `(wallet, pubkey, index)` always yields the same preimage. The signature reveals nothing about the key beyond the preimage itself.

## Why this shape

Considered (and rejected, with reviewer help):

- **`SHA-256(pubkey)` alone** — unsafe; the public key isn't secret (it's on-chain), so a leak would compromise every past and future swap on that key. Signing first roots the preimage in secret key material.
- **New `IArkadeWalletSigner.DerivePreimage` method** — bloats a consumer-facing contract for one subsystem's needs; remote signers would need a `NotSupportedException` path that breaks recovery for them.
- **`IDeterministicPreimageSource` capability sibling + probe on signer** — added one method to `IArkadeWalletSigner` for one feature; pattern not used elsewhere in this codebase.

Sign-and-hash uses the existing `Sign(descriptor, hash)` API. No new abstraction. Works for `Bip39SigningSource`, `NsecSigningSource`, *and* `RemoteTransportSigningSource` uniformly — remote-signed wallets get deterministic preimages and recovery for free *if* the transport signs with `aux_rand=null` (see below).

## Plumbing

- **`BoltzSwapsService`** — three swap-create methods (`CreateReverseSwap`, `CreateBtcToArkSwapAsync`, `CreateArkToBtcSwapAsync`) gain an optional `byte[]? preimage` parameter. `null` falls back to `RandomUtils.GetBytes(32)`, so existing callers + watch-only/no-signer scenarios keep working.
- **`SwapsManagementService`** — single private `DerivePreimageAsync(walletId, descriptor, uint index, ct)` helper fetches a signer for the wallet, builds the bundled message, signs the SHA-256 of it with the descriptor key, and SHA-256s the signature. `GetSignerAsync == null` (watch-only) → random preimage fallback. Wired at all three `Initiate*Swap` call sites with `index: 0`.
- **Restore path (`RestoreSwaps`)** — after `ReconstructContract` resolves the receiver descriptor for a reverse swap, re-derive the candidate preimage (`index: 0`), verify `SHA-256(candidate) == restored.PreimageHash`, and attach to `swap.Metadata[Preimage]` on match. Mismatch leaves it out; `EnrichReverseSwapPreimage` remains the manual path. A future multi-preimage variant should iterate `index=0..MAX` here.

## Remote-signer determinism

Local sources (`Bip39SigningSource`, `NsecSigningSource`) already sign BIP-340 with `aux_rand=null` so signatures are byte-identical per `(key, hash)`. Remote-signer transports must honour the same convention — hardware wallets that randomise `aux_rand` (for side-channel resistance) will silently break recovery for their wallets. The requirement is now explicit in `IRemoteSignerTransport.SignAsync`'s XML doc; recovery on such wallets is a known phase-1 gap.

## Scope

- Reverse + chain swaps only (we control the preimage). Submarine swaps are unaffected — Boltz generates that preimage and reveals it on LN payment.
- `IArkadeWalletSigner` / `IDescriptorSigningSource` surfaces stay unchanged. Only `IRemoteSignerTransport.SignAsync` gets a determinism note in XML (no breaking change).
- Backwards-compatible: existing on-chain VHTLCs generated with random preimages are not recoverable from seed (the random bytes are gone), but they continue to claim the same way they always did — the deterministic path is purely additive.

## Test plan

- [x] `dotnet test NArk.Tests` — 418 passed, 0 failed
- [ ] E2E CI — covered (Boltz swap suite exercises the create + restore paths)

Docs updated: `docs/articles/swaps.md` got a "Deterministic preimages" section under *Restore After Data Loss* describing the bundled message construction, the cross-SDK interop intent, and the remote-signer caveat.

---

## Infra: regtest migration to denigiri (added in this PR)

Bumps the `regtest` submodule to the **arkade-regtest #27 (denigiri)** head, replacing the nigiri binary + shell scripts with the zero-dependency Node orchestrator, and migrates the E2E harness accordingly:

- **build.yml** — Node 20 instead of Go/nigiri-cache; `node regtest/regtest.mjs start --profile boltz,delegate` / `clean`; failure-log container `ark` → `arkd`.
- **Test infra** — Esplora endpoint → mempool's `/api/`; `DockerHelper.BitcoinCli` central helper running `bitcoin-cli -regtest -rpcuser=admin1 -rpcpassword=123` (denigiri's btcpayserver image), routed through every faucet/mine call site; `arkd` container name.

E2E on denigiri: **38 passed / 1 skipped / 0 failed in ~6 min** (vs ~14 on nigiri).

### DIFF
--- a/NArk.Abstractions/Wallets/IRemoteSignerTransport.cs
+++ b/NArk.Abstractions/Wallets/IRemoteSignerTransport.cs
@@ -78,6 +78,16 @@ public interface IRemoteSignerTransport
     /// the descriptor's private key, returning the x-only pubkey alongside the
     /// signature.
     /// </summary>
+    /// <remarks>
+    /// Implementations <b>SHOULD</b> sign with <c>aux_rand</c> set to null (or all-zero) so the
+    /// signature is deterministic per <c>(key, hash)</c>. The SDK relies on this for the
+    /// swap-preimage recovery scheme in <c>SwapsManagementService.DerivePreimageAsync</c>:
+    /// same descriptor + same wallet must produce the same signature, otherwise a restored
+    /// wallet that rediscovers an outstanding swap will re-derive a different preimage and
+    /// fail to claim the VHTLC. Implementations that randomise <c>aux_rand</c> (e.g. for
+    /// side-channel resistance on a hardware signer) break that contract, and remote-signed
+    /// wallets on such transports will not recover outstanding swap preimages from seed.
+    /// </remarks>
     /// <param name="walletId">The wallet whose key signs.</param>
     /// <param name="descriptor">The descriptor identifying the signing key.</param>
     /// <param name="hash">The 32-byte hash to sign.</param>
--- a/NArk.Swaps/Boltz/BoltzSwapsService.cs
+++ b/NArk.Swaps/Boltz/BoltzSwapsService.cs
@@ -73,6 +73,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
 
     public async Task<ReverseSwapResult> CreateReverseSwap(CreateInvoiceParams createInvoiceRequest,
         OutputDescriptor receiver,
+        byte[]? preimage = null,
         CancellationToken cancellationToken = default)
     {
         var extractedReceiver = receiver.Extract();
@@ -80,9 +81,10 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
         // Get operator terms
         var operatorTerms = await clientTransport.GetServerInfoAsync(cancellationToken);
 
-        //TODO: deterministic hash somehow instead?
-        // Generate preimage and compute preimage hash using SHA256 for Boltz
-        var preimage = RandomUtils.GetBytes(32);
+        // Caller-supplied preimage enables deterministic derivation (so restored wallets can
+        // re-derive and claim outstanding swaps); null falls back to random for watch-only or
+        // any other no-signer scenario.
+        preimage ??= RandomUtils.GetBytes(32);
         var preimageHash = Hashes.SHA256(preimage);
 
         // First make the Boltz request to get the swap details including timeout block heights
@@ -167,6 +169,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
     public async Task<ChainSwapResult> CreateBtcToArkSwapAsync(
         long amountSats,
         OutputDescriptor claimDescriptor,
+        byte[]? preimage = null,
         CancellationToken ct = default)
     {
         var operatorTerms = await clientTransport.GetServerInfoAsync(ct);
@@ -174,7 +177,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
         var claimPubKeyHex = (extractedClaim.PubKey?.ToBytes() ?? extractedClaim.XOnlyPubKey.ToBytes())
             .ToHexStringLower();
 
-        var preimage = RandomUtils.GetBytes(32);
+        preimage ??= RandomUtils.GetBytes(32);
         var preimageHash = Hashes.SHA256(preimage);
         var ephemeralKey = new Key();
 
@@ -242,6 +245,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
     public async Task<ChainSwapResult> CreateArkToBtcSwapAsync(
         long amountSats,
         OutputDescriptor refundDescriptor,
+        byte[]? preimage = null,
         CancellationToken ct = default)
     {
         var operatorTerms = await clientTransport.GetServerInfoAsync(ct);
@@ -252,7 +256,7 @@ internal class BoltzSwapService(BoltzClient boltzClient, IClientTransport client
         ).ToLowerInvariant();
 
 
-        var preimage = RandomUtils.GetBytes(32);
+        preimage ??= RandomUtils.GetBytes(32);
         var preimageHash = Hashes.SHA256(preimage);
         var ephemeralKey = new Key();
 
--- a/NArk.Swaps/Services/SwapsManagementService.cs
+++ b/NArk.Swaps/Services/SwapsManagementService.cs
@@ -105,11 +105,14 @@ public class SwapsManagementService : IAsyncDisposable
     // message bundles:
     //   tag:        domain-separates this signature from any other use of the signing key
     //               (versioned so a future scheme bump can coexist on recovery)
-    //   descriptor: scopes the preimage to the specific swap descriptor
-    //   index:      lets the caller derive multiple preimages from the same descriptor;
-    //               always 0 today, but baked into v1 so recovery iteration is forward-
-    //               compatible without a scheme bump
-    // Same (wallet, descriptor, index) → same signature → same preimage, so a restored
+    //   pubkey:     the descriptor's x-only public key — scopes the preimage to the swap
+    //               key. Canonical, unlike descriptor.ToString() (which differs between a
+    //               signing descriptor and the bare receiver descriptor a restore
+    //               reconstructs), so create-time and restore-time derive the same value.
+    //   index:      lets the caller derive multiple preimages from the same key; always 0
+    //               today, but baked into v1 so recovery iteration is forward-compatible
+    //               without a scheme bump
+    // Same (wallet, pubkey, index) → same signature → same preimage, so a restored
     // wallet that rediscovers an outstanding swap via Boltz /v2/swap/restore can re-derive
     // the preimage and claim the VHTLC. Watch-only and remote-signer-less wallets fall
     // through to a random preimage.
@@ -123,6 +126,24 @@ public class SwapsManagementService : IAsyncDisposable
     private const string PreimageTag = "Arkade-Boltz-Preimage-v1";
     private static readonly byte[] PreimageTagBytes = Encoding.UTF8.GetBytes(PreimageTag);
 
+    // Builds the message that gets BIP-340-signed. Format (cross-SDK):
+    // PreimageTag || x-only pubkey (32B) || u32le(index). Anchoring on the canonical
+    // x-only key — not descriptor.ToString(), which is non-canonical and differs
+    // between a signing descriptor and a reconstructed bare receiver descriptor —
+    // keeps create-time and restore-time derivation identical and reproducible by any
+    // Arkade SDK.
+    internal static byte[] BuildPreimageMessage(OutputDescriptor descriptor, uint index)
+    {
+        var keyBytes = OutputDescriptorHelpers.Extract(descriptor).XOnlyPubKey.ToBytes();
+        var indexBytes = BitConverter.GetBytes(index);
+        if (!BitConverter.IsLittleEndian) Array.Reverse(indexBytes); // canonical u32 LE
+        var message = new byte[PreimageTagBytes.Length + keyBytes.Length + indexBytes.Length];
+        Buffer.BlockCopy(PreimageTagBytes, 0, message, 0, PreimageTagBytes.Length);
+        Buffer.BlockCopy(keyBytes, 0, message, PreimageTagBytes.Length, keyBytes.Length);
+        Buffer.BlockCopy(indexBytes, 0, message, PreimageTagBytes.Length + keyBytes.Length, indexBytes.Length);
+        return message;
+    }
+
     private async Task<byte[]> DerivePreimageAsync(
         string walletId, OutputDescriptor descriptor, uint index, CancellationToken cancellationToken)
     {
@@ -130,15 +151,7 @@ public class SwapsManagementService : IAsyncDisposable
         if (signer is null)
             return RandomUtils.GetBytes(32); // watch-only — no entropy floor to draw from
 
-        var descriptorBytes = Encoding.UTF8.GetBytes(descriptor.ToString());
-        var indexBytes = BitConverter.GetBytes(index);
-        if (!BitConverter.IsLittleEndian) Array.Reverse(indexBytes); // canonical u32 LE
-        var message = new byte[PreimageTagBytes.Length + descriptorBytes.Length + indexBytes.Length];
-        Buffer.BlockCopy(PreimageTagBytes, 0, message, 0, PreimageTagBytes.Length);
-        Buffer.BlockCopy(descriptorBytes, 0, message, PreimageTagBytes.Length, descriptorBytes.Length);
-        Buffer.BlockCopy(indexBytes, 0, message, PreimageTagBytes.Length + descriptorBytes.Length, indexBytes.Length);
-        var messageHash = new uint256(SHA256.HashData(message));
-
+        var messageHash = new uint256(SHA256.HashData(BuildPreimageMessage(descriptor, index)));
         var (_, sig) = await signer.Sign(descriptor, messageHash, cancellationToken);
         return SHA256.HashData(sig.ToBytes());
     }
--- a/NArk.Swaps/Services/SwapsManagementService.cs
+++ b/NArk.Swaps/Services/SwapsManagementService.cs
@@ -101,21 +101,42 @@ public class SwapsManagementService : IAsyncDisposable
     public IReadOnlyList<ISwapProvider> Providers => _providers;
 
     // BIP-340 sign+hash gives us a deterministic preimage rooted in the wallet's secret
-    // material without leaking the key (signatures reveal nothing about the key). Same
-    // descriptor + same wallet → same signature → same preimage, so a restored wallet that
-    // rediscovers an outstanding swap via Boltz /v2/swap/restore can re-derive the preimage
-    // and claim the VHTLC. Watch-only and remote-signer-less wallets fall through to a
-    // random preimage (recovery gap, but the swap still works at create time).
-    private static readonly uint256 PreimageSignTagHash =
-        new uint256(SHA256.HashData(Encoding.UTF8.GetBytes("NArk-Boltz-Preimage-v1")));
+    // material without leaking the key (signatures reveal nothing about the key). The signed
+    // message bundles:
+    //   tag:        domain-separates this signature from any other use of the signing key
+    //               (versioned so a future scheme bump can coexist on recovery)
+    //   descriptor: scopes the preimage to the specific swap descriptor
+    //   index:      lets the caller derive multiple preimages from the same descriptor;
+    //               always 0 today, but baked into v1 so recovery iteration is forward-
+    //               compatible without a scheme bump
+    // Same (wallet, descriptor, index) → same signature → same preimage, so a restored
+    // wallet that rediscovers an outstanding swap via Boltz /v2/swap/restore can re-derive
+    // the preimage and claim the VHTLC. Watch-only and remote-signer-less wallets fall
+    // through to a random preimage.
+    //
+    // Local signing sources pass aux_rand=null to BIP-340, so signatures are deterministic.
+    // Remote-signer transports MUST honour the same convention or the preimage will rotate
+    // per call and recovery will silently fail — see IRemoteSignerTransport.SignAsync.
+    private const string PreimageTag = "NArk-Boltz-Preimage-v1";
+    private static readonly byte[] PreimageTagBytes = Encoding.UTF8.GetBytes(PreimageTag);
 
     private async Task<byte[]> DerivePreimageAsync(
-        string walletId, OutputDescriptor descriptor, CancellationToken cancellationToken)
+        string walletId, OutputDescriptor descriptor, uint index, CancellationToken cancellationToken)
     {
         var signer = await _walletProvider.GetSignerAsync(walletId, cancellationToken);
         if (signer is null)
             return RandomUtils.GetBytes(32); // watch-only — no entropy floor to draw from
-        var (_, sig) = await signer.Sign(descriptor, PreimageSignTagHash, cancellationToken);
+
+        var descriptorBytes = Encoding.UTF8.GetBytes(descriptor.ToString());
+        var indexBytes = BitConverter.GetBytes(index);
+        if (!BitConverter.IsLittleEndian) Array.Reverse(indexBytes); // canonical u32 LE
+        var message = new byte[PreimageTagBytes.Length + descriptorBytes.Length + indexBytes.Length];
+        Buffer.BlockCopy(PreimageTagBytes, 0, message, 0, PreimageTagBytes.Length);
+        Buffer.BlockCopy(descriptorBytes, 0, message, PreimageTagBytes.Length, descriptorBytes.Length);
+        Buffer.BlockCopy(indexBytes, 0, message, PreimageTagBytes.Length + descriptorBytes.Length, indexBytes.Length);
+        var messageHash = new uint256(SHA256.HashData(message));
+
+        var (_, sig) = await signer.Sign(descriptor, messageHash, cancellationToken);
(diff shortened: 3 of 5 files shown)


ghost (changes requested):
## Arkana Code Review — `feat(swaps): deterministic Boltz preimages via BIP-340 sign-and-hash`

**Verdict: Request changes.** This is protocol-critical code (swap preimage derivation, VHTLC claiming). The design is sound — sign-and-hash with BIP-340 is the right shape, and reusing the existing `Sign(descriptor, hash)` surface avoids bloating consumer contracts. But there are gaps that need closing before merge.

---

### 🔴 P0 — Must fix

**1. Chain swap recovery is missing**

`DerivePreimageAsync` is wired at all three `Initiate*Swap` call sites (reverse, BtcToArk, ArkToBtc), but the recovery path in `RestoreSwaps` only re-derives for reverse swaps:

- `SwapsManagementService.cs:670` — gated on `restored.IsReverseSwap`
- `MapRestoredSwap` (line ~720) returns `null` for anything other than `"reverse"` or `"submarine"` — chain swaps fall through

This means chain swaps created with deterministic preimages **cannot recover those preimages from seed** via `RestoreSwaps`. The docs (`swaps.md`) claim the scheme covers "reverse and chain swaps" — that's true at creation time but not at recovery time.

Either:
- (a) Extend `MapRestoredSwap` + the recovery branch to handle chain swap types, or
- (b) Scope the docs/PR description to "reverse swaps only for recovery" and leave chain swap recovery as a follow-up

**2. Zero test coverage on `DerivePreimageAsync`**

This is the linchpin of the entire feature. A wrong derivation = silently unclaimable VHTLCs = lost funds. There are no unit tests for:
- **Idempotency**: same `(walletId, descriptor, index)` → same preimage across calls
- **Descriptor isolation**: different descriptors → different preimages
- **Index variation**: different indices → different preimages
- **Watch-only fallback**: `GetSignerAsync == null` → random (non-deterministic) preimage

The method is `private`, but it's testable via `internal` + `InternalsVisibleTo`, or by extracting it to a static helper that takes a signer. Protocol-critical derivation schemes need determinism tests before shipping.

---

### 🟡 P1 — Should fix

**3. `SHOULD` vs `MUST` on `IRemoteSignerTransport.SignAsync`**

`IRemoteSignerTransport.cs:81` says:
> Implementations **SHOULD** sign with `aux_rand` set to null

But `docs/articles/swaps.md` says:
> `IRemoteSignerTransport.SignAsync` **MUST** use `aux_rand=null`

This is a money-losing failure mode (recovery silently fails). The interface contract should say **MUST**. Integrators read the XML doc, not the markdown. If a hardware signer randomises `aux_rand`, every swap preimage is irrecoverable and the user finds out only when they try to restore — too late.

**4. `descriptor.ToString()` stability as a derivation input**

`SwapsManagementService.cs:136` — the preimage is bound to `descriptor.ToString()`. If `OutputDescriptor.ToString()` changes its formatting across SDK versions (whitespace, key encoding, normalization), a wallet restored on a newer SDK version will derive a different message hash → different signature → different preimage → silent claim failure.

Consider either:
- Documenting `OutputDescriptor.ToString()` as a stability invariant that must not change across versions, or
- Using a canonical byte serialization (e.g. the raw pubkey bytes + derivation path) instead of the string representation

**5. Recovery doesn't enrich the VHTLC contract entity**

The recovery path at `SwapsManagementService.cs:680-690` attaches the preimage to `swap.Metadata[Preimage]` only. The existing manual path `EnrichReverseSwapPreimage` (line ~790) also recreates the `VHTLCContract` with the preimage bytes and persists it via `_contractStorage.SaveContract`. The sweeper reads from metadata (`BoltzSwapProvider.cs:1532`), so claiming works, but the contract entity remains un-enriched — inconsistent with the manual path. If any code path ever reads the preimage from the contract rather than swap metadata, it'll silently fail.

---

### 🟢 P2 — Nits / observations

**6. Concatenation without length prefix**

`SwapsManagementService.cs:133-138` — `tag || descriptor || index` concatenation is technically ambiguous (a descriptor ending in bytes that look like a u32 LE index could collide with a shorter descriptor + different index). Practically safe since the tag is constant-length and descriptors are structured strings, but a length-prefixed scheme would be more robust for a v1 spec you're committing to forever.

**7. PR body is stale**

The first commit's scheme was `SHA-256(signer.Sign(descriptor, SHA-256("NArk-Boltz-Preimage-v1")))`. The second commit changed it to include `descriptor || index` in the signed message. The PR body still shows the old scheme. Future readers will be confused.

**8. First-mover across SDKs**

The dotnet-sdk is the first to implement deterministic swap preimages. ts-sdk still uses `randomBytes(32)` at `arkade-swaps.ts:457,2321` and skips unresolvable swaps during restore (`swap-manager.ts:1033`). Worth tracking as a cross-SDK initiative — the derivation spec should be portable so wallets restored across SDK implementations produce the same preimages.

---

### Summary

The cryptographic design is clean: BIP-340 sign-and-hash with domain separation, no new abstraction surface, backward-compatible optional parameters. The main gaps are (1) chain swap recovery not wired, (2) no tests on the derivation function, and (3) the SHOULD/MUST inconsistency on the remote signer contract.

⚠️ **Protocol-critical: requires human review and approval before merge**, even after addressing the above.

ghost (changes requested):
## Arkana Follow-up Review — commit `a05f1fb` (tag rename `NArk` → `Arkade`)

### New commit assessment

The tag rename from `NArk-Boltz-Preimage-v1` to `Arkade-Boltz-Preimage-v1` is correct and well-motivated — SDK-agnostic naming is the right call for cross-SDK preimage recovery interop. Code change is trivial (const + docs), no issues.

---

### Previous findings — status

**None of the P0s or P1s from my first review have been addressed.** Summarizing what's still open:

#### 🔴 P0 — Still blocking

**1. Chain swap recovery still missing**
`MapRestoredSwap` (`SwapsManagementService.cs:710-714`) still returns `null` for anything other than `"reverse"` or `"submarine"`. Chain swaps created with deterministic preimages via `InitiateBtcToArkChainSwap` / `InitiateArkToBtcChainSwap` cannot be recovered from seed. The preimage is derived at create time but lost on restore.

**2. Zero test coverage on `DerivePreimageAsync`**
No test files in any of the three commits. This is the derivation function that determines whether users can recover funds. Needs idempotency, descriptor-isolation, and watch-only-fallback tests at minimum.

#### 🟡 P1 — Still open

**3. `SHOULD` vs `MUST` — `IRemoteSignerTransport.cs:82`**
XML doc still says `SHOULD`. Internal comment at `SwapsManagementService.cs:120` says `MUST`. Docs at `swaps.md` say `MUST`. The XML doc is what integrators read — this should be `MUST`.

**4. `descriptor.ToString()` stability** — no response yet.

**5. Recovery doesn't enrich VHTLC contract entity** — no response yet.

---

Maintaining **request-changes**. The tag rename is good, but P0-1 (chain swap recovery) and P0-2 (tests) are still blockers.

⚠️ **Protocol-critical: requires human review and approval before merge.**

ghost (changes requested):
## Arkana Follow-up Review — commit `b4506f8` (ci: nigiri → denigiri migration)

### New commit assessment

The CI migration from nigiri/Go to denigiri/Node is straightforward. Container rename `ark` → `arkd`, chopsticks → mempool `/api/` base path, submodule bump — all make sense. A few things were missed:

---

### 🟡 P1 — Incomplete migration (stale references)

**1. `SharedDelegationInfrastructure.cs:37` — still references `./regtest/start-env.sh`**

The same error-message fix applied to `SharedArkInfrastructure.cs:29` and `SharedSwapInfrastructure.cs:37` was not applied here. Should be `node regtest/regtest.mjs start --profile boltz,delegate`.

**2. `docs/articles/development.md` — entire file still references old infra**

Lines 28, 32, 39–41, 47 all reference `start-env.sh`, `clean-env.sh`, `stop-env.sh`, and nigiri. This is the developer onboarding doc — if someone follows it they'll get a "command not found". Should be updated to the new `node regtest/regtest.mjs` commands.

**3. `CLAUDE.md:54,58` — references nigiri**

Lines 54 and 58 mention nigiri as a dependency. If the repo's own AI coding assistant instructions are stale, agents will give wrong setup advice.

---

### 🟢 P2 — Nits

**4. `SharedArkInfrastructure.cs:8` — `ChopsticksEndpoint` naming**

The variable is still named `ChopsticksEndpoint` but chopsticks no longer exists (replaced by mempool). The inline comment correctly explains the change, but renaming to e.g. `EsploraEndpoint` or `MempoolEndpoint` would prevent future confusion.

**5. `DockerHelper.cs:71,129` — comments reference "nigiri lnd"**

XML doc on `CreateLndInvoice` (line 71) and `PayLndInvoice` (line 129) still say "nigiri lnd container". Cosmetic, but worth cleaning up in the same pass.

**6. `.env.regtest:3,9` — references `start-env.sh` and nigiri**

Comments in this file still reference the old infra. Minor since it's a config file, but contributes to the overall stale-reference problem.

---

### Previous P0/P1 findings — still open

None of the protocol-critical findings from my first two reviews have been addressed. This commit is infra-only, so that's expected, but they remain blockers:

- **P0-1**: Chain swap recovery not wired in `RestoreSwaps`
- **P0-2**: Zero test coverage on `DerivePreimageAsync`
- **P1-3**: `SHOULD` vs `MUST` on `IRemoteSignerTransport.SignAsync`
- **P1-4**: `descriptor.ToString()` stability as derivation input
- **P1-5**: Recovery doesn't enrich VHTLC contract entity

Maintaining **request-changes**. The CI migration commit needs the stale references cleaned up (P1-1 and P1-2 at minimum), and the protocol-critical items from prior reviews are still outstanding.

⚠️ **Protocol-critical: requires human review and approval before merge.**

ghost (changes requested):
## Arkana Follow-up Review — commits `b21dca2` + `a5f17d5`

### New commits assessment

**Commit 5** (`test(e2e): drive bitcoin-cli with denigiri regtest flags`) — Good infra fix. Centralises the regtest connection flags into `DockerHelper.BitcoinCli` with proper error checking. One minor inconsistency noted below.

**Commit 6** (`fix(swaps): derive preimage from x-only pubkey, not descriptor.ToString()`) — **Excellent fix.** This directly addresses my prior P1-4 finding and is the right design: anchoring on the canonical 32-byte x-only pubkey instead of the non-canonical `descriptor.ToString()`. The message format `tag(24B) || xonly(32B) || u32le(4B)` is fixed-layout with no ambiguity — also resolves the P2-6 length-prefix concern from my first review. The `BuildPreimageMessage` extraction as `internal static` is clean and testable.

The new `PreimageDerivationTests` are well-designed:
- `Message_IsTagPlusXOnlyPubkeyPlusIndexLittleEndian` — pins the wire format
- `Message_IndexEncodedAsUInt32LittleEndian` — validates endianness
- `Message_IsIndependentOfDescriptorStringForm` — the **critical** test: proves HD signing descriptor and bare receiver descriptor produce the same message, which is the exact failure mode this commit fixes

### Prior findings status

| # | Finding | Status |
|---|---------|--------|
| P0-1 | Chain swap recovery missing | ❌ Still open |
| P0-2 | No test coverage on derivation | ✅ Addressed (wire format + descriptor independence tests) |
| P1-3 | SHOULD vs MUST on `aux_rand` | ❌ Still open — `IRemoteSignerTransport.cs:82` says SHOULD, `swaps.md:128` says MUST |
| P1-4 | `descriptor.ToString()` stability | ✅ Fixed — switched to x-only pubkey |
| P1-5 | Recovery doesn't enrich VHTLC contract entity | ❌ Still open |
| P2-6 | Concatenation without length prefix | ✅ Resolved — fixed-layout format now |
| P2-7 | Stale PR body | ❌ Still open |

### New findings in these commits

**1. `MineBlocks` inconsistency** (minor)

`DockerHelper.cs` — `MineBlocks` still routes through `Exec()` (with the spread `BitcoinCliArgs`) instead of using the new `BitcoinCli()` helper:

```csharp
public static async Task MineBlocks(int count = 20, CancellationToken ct = default)
    => await Exec("bitcoin", [.. BitcoinCliArgs, "-generate", count.ToString()], ct);
```

`Exec` doesn't check exit codes or trim output. `BitcoinCli` does both. A silent `MineBlocks` failure would make tests hang waiting for confirmations rather than fail fast. Should use `BitcoinCli(["-generate", count.ToString()], ct)` for consistency.

### Remaining blockers

The P0-1 (chain swap recovery) is still the primary blocker. The `MapRestoredSwap` method only handles `"reverse"` and `"submarine"` types — chain swaps that were created with deterministic preimages cannot recover those preimages via `RestoreSwaps`. The commit message for `a5f17d5` specifically mentions the restore scenario, and the new x-only-pubkey scheme makes it *possible* to recover correctly, but the `RestoreSwaps` code path doesn't exercise it for chain swaps.

The P1-3 (`SHOULD` → `MUST` on `aux_rand`) is a one-word fix with outsized impact — integrators read the XML doc, not the markdown. A hardware signer that randomises `aux_rand` silently breaks every swap recovery. This should be `MUST` in the interface contract.

Maintaining **request-changes** on P0-1 and P1-3. The preimage derivation scheme itself is now solid — the x-only-pubkey anchor is the right call.

ghost (changes requested):
## Arkana Follow-up Review — commits `b21dca20`, `a5f17d54`, `3bbcd18e`

Three new commits since my last review. The big one is `a5f17d54` which fixes the `descriptor.ToString()` instability I flagged as P1-4.

---

### New commits assessment

**`a5f17d54` — fix(swaps): derive preimage from x-only pubkey** ✅ **Good fix**

This directly addresses my prior P1-4 finding. The change is correct:
- `BuildPreimageMessage` now anchors on `Extract(descriptor).XOnlyPubKey.ToBytes()` (canonical 32-byte x-only key) instead of `descriptor.ToString()` (non-canonical, differs between HD signing descriptor and bare receiver descriptor at restore time).
- Extracted as `internal static` pure function — testable, reviewable, and reusable.
- `PreimageDerivationTests.cs` pins the wire format with 3 tests:
  - `Message_IsTagPlusXOnlyPubkeyPlusIndexLittleEndian` — byte-level format correctness
  - `Message_IndexEncodedAsUInt32LittleEndian` — index encoding
  - `Message_IsIndependentOfDescriptorStringForm` — **the critical invariant**: HD signing descriptor vs bare receiver descriptor → same message

One subtlety worth noting: the test's HD descriptor uses `Extract()` which resolves the wildcard at position 0 via `hd.GetPubKey(0, _ => null, out _)`. This works because both create-time (`GetNextSigningDescriptor`) and restore-time (`contract.Receiver`) ultimately resolve to the same x-only key for the swap. The test correctly demonstrates this invariant.

**`b21dca20` — test(e2e): bitcoin-cli denigiri connection flags** ✅ **Clean**

Centralizes `bitcoin-cli` invocations behind `DockerHelper.BitcoinCli()` with proper `-regtest -rpcuser=admin1 -rpcpassword=123` flags. Good dedup. Fixes 8 test failures from the denigiri migration.

Minor nit: `MineBlocks` at `DockerHelper.cs:59-60` still uses `Exec()` directly with `BitcoinCliArgs` instead of routing through `BitcoinCli()`. `Exec()` doesn't throw on non-zero exit, so mine failures are silently swallowed. Should either use `BitcoinCli(["-generate", count.ToString()], ct)` for consistency, or document that mine is intentionally best-effort.

**`3bbcd18e` — docs(swaps): update scheme description** ✅ **Correct**

Docs now accurately describe `xonly_pubkey(32)` instead of `descriptor.ToString()`. Cross-SDK interop rationale is clear.

---

### Previously flagged items — status

| # | Finding | Status |
|---|---------|--------|
| P0-1 | Chain swap recovery missing (`MapRestoredSwap` returns null for chain types, `RestoreSwaps:691` gated on `IsReverseSwap`) | **Still open** |
| P0-2 | Zero test coverage on derivation | **Addressed** — `PreimageDerivationTests` covers `BuildPreimageMessage`. End-to-end `DerivePreimageAsync` tests (idempotency through actual signing, watch-only fallback) are still missing but the critical format invariant is now pinned. |
| P1-3 | `SHOULD` vs `MUST` on `IRemoteSignerTransport.SignAsync:82` `aux_rand` | **Still open** — still says SHOULD, docs say MUST. This is a money-losing mismatch. |
| P1-4 | `descriptor.ToString()` stability | **Fixed** ✅ |
| P1-5 | Recovery doesn't enrich VHTLC contract entity | **Still open** |

---

### Verdict

The x-only pubkey fix is solid and well-tested — that was the biggest technical risk. The CI/test infrastructure changes are clean.

Maintaining **request-changes** because P0-1 (chain swap recovery) and P1-3 (SHOULD→MUST on aux_rand) are still outstanding. P0-1 is the more important one: chain swaps created with deterministic preimages can't recover them from seed.

🤖 *Arkana iterative review — focused on commits since `b4506f8`*

ghost (changes requested):
## Arkana Follow-up Review — rebase with major new code (`503ed10`..`f7577ba`)

The branch was rebased and pulled in ~3,600 new lines beyond the preimage commits I previously reviewed: BoltzSwapProvider partial-class refactor, `BoltzOperationClassifier`, `WalletRecoveryService`, expanded E2E chain-swap tests, `IndexerVtxoDiscoveryProvider` rewrite, and a `BatchManagementService` race fix. Reviewed all of it.

---

### Previously flagged items — status

| # | Finding | Status |
|---|---------|--------|
| P0-1 | Chain swap recovery missing in `MapRestoredSwap` | **❌ Still open** — line 763-769: only `"reverse"` and `"submarine"` are mapped; chain types fall through to `null`. Preimage re-derivation at line 733 still gated on `restored.IsReverseSwap`. |
| P0-2 | No test coverage on derivation | ✅ Fixed — `PreimageDerivationTests` pins wire format + descriptor independence |
| P1-3 | `SHOULD` vs `MUST` on `aux_rand` | **❌ Still open** — `IRemoteSignerTransport.cs:82` says SHOULD; `swaps.md` says MUST |
| P1-4 | `descriptor.ToString()` stability | ✅ Fixed — switched to x-only pubkey |
| P1-5 | Recovery doesn't enrich VHTLC contract entity | **❌ Still open** |

---

### New code — protocol-critical findings

**🟢 APPROVE: ARK→BTC VHTLC address verification** (`BoltzSwapsService.cs:~281`)

The `CreateArkToBtcSwapAsync` path now independently reconstructs the VHTLC contract from Boltz's response and validates `computedAddress == lockupDetails.LockupAddress`. This prevents a malicious Boltz from providing a lockup address that doesn't match the expected VHTLC — a real fund-theft vector the old code was vulnerable to. Excellent hardening.

**🟢 APPROVE: BatchManagementService race fix** (`BatchManagementService.cs:~170`)

The reconciliation on `StreamStartedEvent` — re-subscribing all active intent topics — fixes a real race where an intent reaching `WaitingForBatch` during the gRPC stream-connect window had its topic subscription silently dropped, causing a permanent settle stall. The `UpdateStreamTopics` add is idempotent, so re-adding is safe. Correct fix, well-commented.

**🟢 APPROVE: BoltzOperationClassifier** (`BoltzOperationClassifier.cs`)

Clean separation of the action-routing decision from inline if/else chains. Test coverage in `BoltzOperationClassifierTests` is exemplary — every Boltz status × swap type combination is pinned. The narrowing of chain swap refund triggers to `swap.expired` only (with `transaction.lockupFailed` routed through renegotiation first) is correct per Boltz's protocol.

**🟢 APPROVE: IndexerVtxoDiscoveryProvider rewrite** (`IndexerVtxoDiscoveryProvider.cs`)

The `Lazy<Task>` → `SemaphoreSlim + nullable` fix eliminates a real bug where a faulted first `GetServerInfoAsync` call permanently poisoned all subsequent probes. The cross-product candidate generation (`{current + deprecated signers} × {current + legacy delay}`) with mainnet-only legacy-delay probing matches ts-sdk behavior. Correct.

**🟢 APPROVE: BoltzSwapProvider partial-class split**

Verified all methods are preserved. The `_spendingService` field removal is correct (dead code). The submarine refund destination caching via `GetOrDeriveRefundDestinationAsync` fixes the orphan contract leak on retries. Status mapping change (operational statuses → `null` instead of `Failed` at restore time) is correct — lets the poll loop take corrective action.

---

### New code — issues to address

**🟡 P1-NEW-1: No delegate contract test coverage** (`IndexerVtxoDiscoveryProviderTests.cs`)

`BuildCandidates` has a delegate-contract branch that generates `ArkDelegateContract` variants for each configured delegate descriptor. None of the 4 tests pass a `RecoveryDelegateConfig`, so that code path is completely untested. If the delegate contract construction is wrong, recovery will silently miss delegated VTXOs — funds stuck until manual intervention.

**🟢 P2-NEW-1: SQL interpolation in DockerHelper** (`DockerHelper.cs`)

`SetBtcToArkChainSwapExpiredWithLockup` and `SetArkToBtcChainSwapExpiredWithLockup` use direct string interpolation into SQL (`$"... WHERE id = '{swapId}'"`) while `TrySetBoltzSwapStatus` correctly uses psql `-v` variable binding. Test-only, zero practical risk (swap IDs are alphanumeric), but worth making consistent.

---

### Remaining blockers

**P0-1 is still the primary blocker.** Chain swaps are created with deterministic preimages via `InitiateBtcToArkChainSwap` / `InitiateArkToBtcChainSwap`, but `MapRestoredSwap` returns `null` for chain swap types, and the preimage re-derivation is gated on `IsReverseSwap`. A wallet restored from seed that had outstanding chain swaps will silently skip them — the user discovers the problem when their VHTLC expires unclaimed.

**P1-3 remains a one-word fix** (`SHOULD` → `MUST` at `IRemoteSignerTransport.cs:82`) with outsized impact — integrators read the XML doc, not the markdown.

Maintaining **request-changes** on P0-1 and P1-3.

⚠️ **Protocol-critical: requires human review and approval before merge.**

🤖 *Arkana iterative review — focused on new code since rebase*

PR #118, opened 2026-06-01
fix(e2e): close batch-topic subscription race + serialize concurrent submarine swap creation

Kukks:
Hardens the dotnet-sdk E2E suite so master goes (and stays) green. Two independent flakes, two fixes.

## 1. Batch-topic subscription race (production fix) — `BatchManagementService`

**Symptom:** `CanStartUnilateralExitForSettledVtxo` intermittently times out after 2 min waiting on `BatchSucceeded`. arkd logs show the intent registered and was then **re-proposed in every batch round for 6+ minutes, never finalizing**.

**Root cause:** When an intent reaches `WaitingForBatch`, `OnIntentChanged` subscribes the event stream to its topics via fire-and-forget `UpdateTopicsAsync`. That call **silently skips when `_streamId` is still null** (stream-connect handshake hasn''t returned `StreamStartedEvent` yet) and never retries. An intent that crosses into `WaitingForBatch` inside that connect window is never subscribed → never receives its `BatchStartedEvent` → never signs → arkd re-proposes it forever. A permanent settle stall, timing-dependent (hence flaky), and able to strand **any** settle test.

**Fix:** on `StreamStartedEvent`, reconcile by re-subscribing every active intent''s topics. `UpdateStreamTopics` add is idempotent (server-side set union) and already exception-guarded, so re-adding existing topics is a no-op — the reconcile only adds topics stranded during the connect window. Internal bugfix, no public API change.

## 2. Concurrent submarine swap creation race (test fix) — `SwapManagementServiceTests`

**Symptom:** `ConcurrentSubmarineSwapsBothComplete` failed master at `6b69b9e` (#114) with `HttpRequestException: {"error":"could not serialize access due to read/write dependencies among transactions"}`.

**Root cause:** two parallel `InitiateSubmarineSwap` calls POST Boltz `/v2/swap/submarine` simultaneously; Boltz''s SERIALIZABLE Postgres backend aborts one with SQLSTATE 40001.

**Fix:** gate only the creation call with a shared `SemaphoreSlim` so both swaps still poll and settle concurrently but don''t collide on Boltz''s swap-insert transaction.

## Test plan

- [ ] `ConcurrentSubmarineSwapsBothComplete` green
- [ ] `CanStartUnilateralExitForSettledVtxo` green
- [ ] Full E2E suite green across **5 sequential runs** to confirm both races are gone

## Notes

- Verified locally only by compile (E2E needs nigiri/arkd/boltz); behavioural verification is via CI.
- Sibling `SubmarineAndReverseSwapsCompleteInParallel` already wraps its reverse-swap creation in `RetryWithSettle`; left as-is.

### DIFF
--- a/NArk.Core/Services/BatchManagementService.cs
+++ b/NArk.Core/Services/BatchManagementService.cs
@@ -167,6 +167,18 @@ public class BatchManagementService(
             case StreamStartedEvent streamStarted:
                 _streamId = streamStarted.StreamId;
                 logger?.LogDebug("Event stream started with ID {StreamId}", _streamId);
+                // An intent that reached WaitingForBatch during the stream-connect
+                // window (before _streamId was set) had its OnIntentChanged topic
+                // subscription silently skipped by UpdateTopicsAsync's null-stream
+                // guard, with no retry. Left unsubscribed, the intent never receives
+                // its BatchStartedEvent, never participates in signing, and gets
+                // re-proposed every batch round forever (a permanent settle stall).
+                // Now that the stream is live, reconcile by re-subscribing every
+                // active intent's topics. UpdateStreamTopics add is idempotent, so
+                // re-adding already-subscribed topics is a no-op.
+                var reconcileTopics = GetAllTopics();
+                if (reconcileTopics.Length > 0)
+                    await UpdateTopicsAsync(addTopics: reconcileTopics);
                 break;
 
             case BatchStartedEvent batchStarted:
--- a/NArk.Tests.End2End/SwapManagementServiceTests.cs
+++ b/NArk.Tests.End2End/SwapManagementServiceTests.cs
@@ -407,6 +407,14 @@ public class SwapManagementServiceTests
         var settledA = new TaskCompletionSource();
         var settledB = new TaskCompletionSource();
 
+        // Boltz serializes swap creation through its Postgres backend under
+        // SERIALIZABLE isolation; two CreateSubmarineSwap POSTs racing in
+        // parallel abort one another with SQLSTATE 40001 ("could not serialize
+        // access due to read/write dependencies among transactions"). Gate just
+        // the creation call so both swaps still poll and settle concurrently
+        // afterwards but don't collide on Boltz's swap-insert transaction.
+        using var createGate = new SemaphoreSlim(1, 1);
+
         async Task<string> RunSwapAsync(
             (ISafetyService safetyService, InMemoryWalletProvider walletProvider, string walletIdentifier,
                 IVtxoStorage vtxoStorage, ContractService contractService, IContractStorage contracts,
@@ -453,7 +461,17 @@ public class SwapManagementServiceTests
 
             var invoice = BOLT11PaymentRequest.Parse(
                 await DockerHelper.CreateLndInvoice(invoiceSats, expirySecs: 0), Network.RegTest);
-            var swapId = await swapMgr.InitiateSubmarineSwap(prereq.walletIdentifier, invoice, autoPay: true, token);
+
+            string swapId;
+            await createGate.WaitAsync(token);
+            try
+            {
+                swapId = await swapMgr.InitiateSubmarineSwap(prereq.walletIdentifier, invoice, autoPay: true, token);
+            }
+            finally
+            {
+                createGate.Release();
+            }
 
             await settledTcs.Task.WaitAsync(TimeSpan.FromMinutes(5), token);
             await swapMgr.DisposeAsync();


ghost (approved):
## Review: ✅ APPROVE

**Scope**: Test-only fix — single file (`NArk.Tests.End2End/SwapManagementServiceTests.cs`), zero production code changes.

### Analysis

The fix is correct and minimal:

1. **Root cause is legitimate.** Boltz runs Postgres under `SERIALIZABLE` isolation. Two concurrent `POST /v2/swap/submarine` calls race on the swap-insert transaction, one gets aborted with `SQLSTATE 40001`. This is a Boltz-side serialization error, not an SDK bug.

2. **Fix is appropriate.** A `SemaphoreSlim(1,1)` gates only the `CreateSubmarineSwap` call so the two swaps don't collide on Boltz's DB transaction. The rest of the test (polling, websocket subscriptions, settlement) still runs concurrently — the test continues to exercise the concurrency paths it was designed for.

3. **Implementation is clean:**
   - `using var createGate` ensures disposal at method exit.
   - `WaitAsync(token)` correctly threads the cancellation token.
   - `try/finally` ensures the semaphore is always released, even if `InitiateSubmarineSwap` throws.
   - Scoped to the local function — no class-level static state that could leak between tests.

4. **No cross-repo impact.** No public API, type, or interface changes. Test-only.

5. **Sibling test noted.** PR description correctly identifies `SubmarineAndReverseSwapsCompleteInParallel` as a potential candidate for the same treatment (it creates a submarine + reverse in parallel), but its reverse-swap path already uses `RetryWithSettle` and the observed conflict was submarine-vs-submarine specifically. Reasonable to defer.

No protocol-critical code touched. No security concerns. LGTM.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (approved):
## Iterative Review — new commit `226d5dd` (batch topic reconciliation)

Previous review approved the test-only semaphore fix. This review covers the new **production code** in `BatchManagementService.cs`.

### Analysis: topic reconciliation on `StreamStartedEvent`

**Bug is real and well-diagnosed.** The race window:
1. Intent reaches `WaitingForBatch` → `OnIntentChanged` fires → fire-and-forget `UpdateTopicsAsync`
2. `UpdateTopicsAsync` checks `_streamId is null` → silently returns (line 98-101) with **no retry**
3. Intent is in `_activeIntents` but never subscribed → never gets `BatchStartedEvent` → permanent settle stall

**Fix is correct:**
- `ProcessEventAsync` → `StreamStartedEvent` case (line 170-179): after setting `_streamId`, calls `GetAllTopics()` to collect every active intent's topics, then `await UpdateTopicsAsync(addTopics: reconcileTopics)`.
- `GetAllTopics()` reads `_activeIntents` (ConcurrentDictionary, snapshot enumeration) — safe.
- `UpdateTopicsAsync` is serialized by `_topicUpdateSemaphore` — no concurrent mutation of server-side topic set.
- Server-side `UpdateStreamTopics` add is idempotent (set union) — re-adding already-subscribed topics is a no-op.
- The `await` blocks the event loop, ensuring topics are subscribed **before** any `BatchStartedEvent` is processed. This is the right ordering.

**Thread-safety verified for all race orderings:**
- Intent added to `_activeIntents` *before* `_streamId` set → `OnIntentChanged`'s `UpdateTopicsAsync` skips → reconcile picks it up. ✅
- Intent added *during* `GetAllTopics()` snapshot → `_streamId` is already set, so `OnIntentChanged`'s own `UpdateTopicsAsync` succeeds independently. ✅
- Intent added *after* reconcile completes → `_streamId` is set, normal path works. ✅

**Reconnection is sound.** On disconnect, `_streamId = null` (catch blocks in `RunSingleEventStreamAsync`). On reconnect, initial topics are passed to `GetEventStreamAsync`, then `StreamStartedEvent` re-reconciles. No stale subscription state leaks across connections.

**One minor observation (not blocking):** If `UpdateStreamTopicsAsync` throws during reconciliation, the exception is caught and logged as a warning (line 106-108), and the stranded topics remain unsubscribed until the next stream reconnect. This matches the existing fire-and-forget error handling — acceptable, but a retry-once on the reconcile path could improve resilience. Not a blocker.

### Verdict: ✅ APPROVE

Fix is minimal, correct, and addresses a real settle-stall race that affects **any** intent crossing `WaitingForBatch` during the stream-connect window. No public API changes, no cross-repo impact.

⚠️ **Human review requested** — this is round-lifecycle code (`BatchManagementService` intent subscription / batch participation). Per protocol-critical review policy, a human must also sign off.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #124, opened 2026-06-08
feat(core):  EARS Coin Selector

d4rp4t:
Introduces EARS, a multi-strategy coin selector for Arkade VTXOs that treats expiry as a first-class constraint.

Problem

The existing DefaultCoinSelector is a plain greedy selector with no awareness of VTXO expiry. When mixing VTXOs from different batches in a single output, the output inherits the youngest (shortest) expiry, silently shortening the effective lifetime of all selected coins. EARS prevents this by default and makes the tradeoff explicit when mixing is unavoidable.

What's new

Four selection strategies, run in parallel with the best result picked by waste metric (change amount + input count):

- ExpiryFirst - deterministic greedy, earliest-expiry bucket first. O(n), baseline.
- BnB - branch-and-bound DFS with suffix-sum pruning, targets exact match (change = 0). Skips groups larger than MaxBnBInputs (default 12).
- RGLI - randomized greedy with local improvement (remove-smallest + single-swap). Multiple random seeds, converges to near-optimal.
- SRD - single random draw, one shuffle + one greedy pass. Privacy fallback.

Sliding window bucketing (ExpiryWindowBlocks = 144, ~24h): batches expiring within 144 blocks of each other are treated as one bucket, so close-expiry VTXOs are spent and renewed together rather than split across separate selection rounds.

ExpiryMixedFallback flag on SelectionResult marks results that mix expiry groups, signalling the caller that a batch join is advisable.

Design decisions

- ExpiryGroup is uint ExpiresAtHeight - exact block height, not a date string
- ICoinSelectionStrategy receives pre-built IReadOnlyList<ExpiryBucket> — bucketing happens once in CoinSelectionEngine, not duplicated per strategy
- EARSCoinSelector implements ICoinSelector - drop-in replacement for DefaultCoinSelector

Test plan

- [ ] 38 unit tests covering all four strategies + bucket building
- [ ] ExpiryFirst: exact match, greedy accumulation, dust rejection, multi-group fallback, mixing flag
- [ ] RGLI: local improvement removes excess coin, swap finds exact match — verified analytically (any seed converges, no retry loop needed)
- [ ] BnB: exact match, minimal waste, dust rejection, MaxBnBInputs limit, earliest-expiry preference
- [ ] SRD: coverage, dust, mixing fallback
- [ ] BuildBuckets: window clustering, gap splitting, ordering


ghost (changes requested):
## EARS Coin Selector — Aggressive Review

Good design direction. Expiry-aware selection is a real gap. But there are blocking issues before this can merge.

---

### 🔴 P0 — Asset selection is silently broken

**`EARSCoinSelector.cs:169-181`** — The asset-aware `SelectCoins` overload accepts `assetRequirements`, passes them into `SelectionContext.AssetRequirements`, and then **no strategy ever reads that field**. All four strategies select purely on BTC value.

This means if a caller uses EARS for an asset transfer, it will select coins that satisfy the BTC target but may contain zero of the required asset. The existing `DefaultCoinSelector` has ~40 lines of explicit asset-first logic (select asset-carrying coins, then fill remaining BTC). EARS has none.

**Impact:** Silent wrong-coin selection for any asset send. Money/asset loss vector.

**Fix:** Either (a) replicate the asset-first selection in `EARSCoinSelector` before delegating to the engine (like `DefaultCoinSelector` does), or (b) have the engine/strategies consume `AssetRequirements`. Option (a) is simpler and keeps strategies BTC-only.

---

### 🔴 P1 — `ExpiryGroup=0` bucketing bug defeats the entire feature

**`CoinSelectionEngine.cs:82`** — The bucketing condition:
```csharp
else if (windowStart == 0u || coin.ExpiryGroup - windowStart <= windowBlocks)
```

When the first coin has `ExpiryGroup == 0` (i.e., `ExpiresAtHeight` is null — boarding UTXOs, unconfirmed VTXOs, etc.), `windowStart` is set to `0u`. The `windowStart == 0u` check then evaluates to `true` for **every subsequent coin**, regardless of its actual expiry height. All coins end up in one bucket.

This completely defeats expiry-aware selection whenever any no-expiry coin is present.

**Fix:** Handle `ExpiryGroup == 0` coins separately — put them in their own bucket, or filter them out and add them back after bucketing. The window comparison should only apply to non-zero expiry groups.

---

### 🟡 P2 — SRD returns null on first sufficient bucket instead of trying others

**`SingleRandomDrawStrategy.cs:36-42`** (diff lines 695-702):
```csharp
foreach (var bucket in buckets)
{
    if (bucket.TotalValue < context.TargetAmount)
        continue;
    var shuffled = bucket.Coins.OrderBy(_ => Random.Shared.Next()).ToList();
    return Greedy(shuffled, context, bucket.ExpiryGroup, expiryMixed: false);
}
```

If `Greedy` returns `null` (e.g., because the random draw produced sub-dust change), the method returns `null` instead of trying the next bucket. Should `continue` on null result, not `return`.

---

### 🟡 P3 — Waste metric ignores input count / fee cost

All strategies set `Waste: change`. The engine picks the result with lowest waste, then breaks ties on input count (`ThenBy(r => r.SelectedCoins.Count)`).

Real waste = `change + (inputCount × costPerInput)`. Without the input cost term, a 10-input exact match (waste=0) always beats a 1-input selection with 100 sat change (waste=100), even though the 10-input version costs ~1500 more sats in fees. This systematically biases toward too many inputs.

At minimum, add a `costPerInput` parameter to `CoinSelectionPolicy` and include it in waste. The `Weight` field on `CoinCandidate` exists but is unused — this is the natural place.

---

### 🟡 P4 — BnB has no mixing fallback (inconsistent with other strategies)

`BranchAndBoundStrategy.TrySelect` iterates single buckets but never attempts cross-bucket selection when `AllowExpiryMixingFallback` is true. The other three strategies all have mixing fallback paths. If BnB is the only strategy that finds a solution in some edge case, the lack of fallback means the engine throws `NotEnoughFundsException` when funds are actually available.

---

### 🟡 P5 — Hardcoded dust threshold in BuildCandidates

**`EARSCoinSelector.cs:189`**: `IsDustProne: c.TxOut.Value < 546` — hardcodes 546 instead of using the `dustThreshold` parameter. Also, `IsDustProne` is never read by any strategy. Either use it or remove it.

---

### Minor / Cleanup

| Location | Issue |
|---|---|
| `CoinSelectionPolicy.cs`, `SelectionContext.cs`, `SelectionResult.cs`, `SelectionStrategy.cs` | Missing newline at EOF |
| All files | Namespace `EARCoinSelector` doesn't match folder name `EARS` |
| `CoinCandidate.Weight` | Computed as `ScriptPubKey.Length` (output script length) — not the input witness weight. Unused, but misleading if someone starts using it |
| PR description | Says strategies "run in parallel" — they run sequentially in `foreach`. Not wrong, but misleading |
| `EARSCoinSelector.BuildContext` | `AllowExpiryMixing` is hardcoded to `false` but `CoinSelectionPolicy.AllowExpiryMixingFallback` defaults to `true` — these are two different flags with unclear interaction. `context.AllowExpiryMixing` is never read by any strategy |

---

### What's good

- Clean strategy pattern with single responsibility per strategy
- `BuildBuckets` as shared infrastructure (bucketing once, not per-strategy)
- BnB suffix-sum pruning is correct and efficient
- RGLI local improvement (remove-smallest + swap) converges well
- Comprehensive test suite — 38 tests covering all strategies and edge cases
- `ExpiryMixedFallback` flag on results is a smart API choice

---

### Verdict

**Request changes.** P0 (asset selection) and P1 (ExpiryGroup=0 bucketing) are blocking — both are silent correctness bugs that could cause wrong coin selection. P2-P5 should be addressed but aren't individually blocking.

⚠️ **Protocol-critical flag:** This code selects VTXOs for spending. Wrong selection → wrong coins in a round → potential fund loss. Requesting human review regardless of code fixes.

ghost (changes requested):
## Follow-up Review — Commits `86c73477`, `d991c38b`, `b88fe0ba`

Previous P0-P5 findings are all addressed. Asset selection, bucketing, SRD null-return, waste metric, BnB fallback, and hardcoded dust are fixed. Good work — the overall structure is much stronger now.

Three remaining issues and one test-coverage gap:

---

### 🟡 P2 — Inconsistent waste in fallback paths (ExpiryFirst + SRD)

**`ExpiryFirstStrategy.cs:84`** and **`SingleRandomDrawStrategy.cs:85`** — The fallback return (loop exhausted all coins, sub-dust change allowed) uses `Waste: finalChange` instead of `CoinSelectionEngine.ComputeWaste(finalChange, selected.Count, policy)`.

The early-return paths at **`ExpiryFirstStrategy.cs:64`** and **`SingleRandomDrawStrategy.cs:66`** correctly use `ComputeWaste`. The fallback path doesn't. When the engine picks between results via `MinBy(r => r.Waste)`, a fallback result that used 5 inputs but only recorded `Waste: 200 sats` will beat a non-fallback result with 1 input and `Waste: 268 sats (200 + 68)` — even though the 5-input result is objectively worse.

**Fix:** Replace `Waste: finalChange` → `Waste: CoinSelectionEngine.ComputeWaste(finalChange, selected.Count, policy)` in both files.

---

### 🟡 P3 — Merged result doesn't recalculate Waste

**`CoinSelectionEngine.cs:64-69`** — After merging asset coins with BTC coins (Phase 4), `Change` is recalculated to reflect the full set but `Waste` is inherited from `best` which only accounts for BTC-only inputs:

```csharp
return best with
{
    SelectedCoins = merged,
    TotalValue = mergedTotal,
    Change = mergedTotal - context.TargetAmount,
    // Waste is not updated — still reflects only BTC selection
};
```

If 3 asset coins were reserved in Phase 1 and BnB found a 1-input BTC solution with waste=68, the merged 4-input result reports waste=68 instead of ~340. Not a correctness bug today (waste isn't consumed downstream), but it's a landmine for anyone who reads `SelectionResult.Waste` for logging or fee estimation.

**Fix:** Add `Waste: CoinSelectionEngine.ComputeWaste(mergedTotal - context.TargetAmount, merged.Count, policy)` to the `with` expression.

---

### 🟡 P4 — Greedy multi-asset selection can fail when valid assignment exists

**`CoinSelectionEngine.cs:82-104`** — `SelectAssetCoins` iterates requirements in order and greedily claims coins. A coin carrying multiple assets is locked to the first requirement that encounters it, even if reassigning it would satisfy all requirements.

Example:
- Coin A: [AssetX: 5, AssetY: 10], Coin B: [AssetX: 10]
- Requirements: [AssetX: 10, AssetY: 10]
- Greedy: claims A for X (total=5), claims B for X (total=15) ✓. Then Y: A already reserved → throws `NotEnoughFundsException`
- Optimal: B→X, A→Y → both satisfied

This is a known limitation of greedy allocation but worth a `// TODO` comment so future maintainers know it's intentional, not accidental. For single-asset sends (the 99% case) this is fine.

---

### 🔵 Missing: zero test coverage for `SelectAssetCoins` / asset-aware engine path

The new `SelectAssetCoins` method and the 4-phase engine flow (reserve assets → check if BTC covered → run strategies on remainder → merge) have no tests. All existing engine tests are for `BuildBuckets` only. Suggest adding at minimum:
- Asset coins reserved correctly + BTC contribution subtracted
- Asset-only result when asset coins cover BTC target
- `NotEnoughFundsException` on insufficient asset balance
- Merged result has correct `SelectedCoins`, `TotalValue`, `Change`

---

### Nit: dead parameter

**`BranchAndBoundStrategy.cs:118`** — `PickBetter` accepts `CoinSelectionPolicy policy` but never reads it. Remove or use.

---

**Summary:** The critical bugs are fixed. What's left is two waste-calculation inconsistencies (P2, P3) and test coverage. P2 can cause suboptimal coin selection in edge cases. P4 is a known greedy limitation worth documenting. None of these are blockers if you're comfortable with the trade-offs, but P2 is a one-line fix in two files — worth doing.

⚠️ **Protocol-critical flag:** This code selects VTXOs for spending. While the fixes look correct, human review is required before merge.

ghost (changes requested):
## Follow-up Review — Commits `86c73477`, `d991c38b`, `b88fe0ba`

Previous P0-P5 findings are all addressed. Asset selection, bucketing, SRD null-return, waste metric, BnB fallback, and hardcoded dust are fixed. Good work — the overall structure is much stronger now.

Three remaining issues and one test-coverage gap:

---

### 🟡 P2 — Inconsistent waste in fallback paths (ExpiryFirst + SRD)

**`ExpiryFirstStrategy.cs:84`** and **`SingleRandomDrawStrategy.cs:85`** — The fallback return (loop exhausted all coins, sub-dust change allowed) uses `Waste: finalChange` instead of `CoinSelectionEngine.ComputeWaste(finalChange, selected.Count, policy)`.

The early-return paths at **`ExpiryFirstStrategy.cs:64`** and **`SingleRandomDrawStrategy.cs:66`** correctly use `ComputeWaste`. The fallback path doesn't. When the engine picks between results via `MinBy(r => r.Waste)`, a fallback result that used 5 inputs but only recorded `Waste: 200 sats` will beat a non-fallback result with 1 input and `Waste: 268 sats (200 + 68)` — even though the 5-input result is objectively worse.

**Fix:** Replace `Waste: finalChange` → `Waste: CoinSelectionEngine.ComputeWaste(finalChange, selected.Count, policy)` in both files.

---

### 🟡 P3 — Merged result doesn't recalculate Waste

**`CoinSelectionEngine.cs:64-69`** — After merging asset coins with BTC coins (Phase 4), `Change` is recalculated to reflect the full set but `Waste` is inherited from `best` which only accounts for BTC-only inputs:

```csharp
return best with
{
    SelectedCoins = merged,
    TotalValue = mergedTotal,
    Change = mergedTotal - context.TargetAmount,
    // Waste is not updated — still reflects only BTC selection
};
```

If 3 asset coins were reserved in Phase 1 and BnB found a 1-input BTC solution with waste=68, the merged 4-input result reports waste=68 instead of ~340. Not a correctness bug today (waste isn't consumed downstream), but it's a landmine for anyone who reads `SelectionResult.Waste` for logging or fee estimation.

**Fix:** Add `Waste: CoinSelectionEngine.ComputeWaste(mergedTotal - context.TargetAmount, merged.Count, policy)` to the `with` expression.

---

### 🟡 P4 — Greedy multi-asset selection can fail when valid assignment exists

**`CoinSelectionEngine.cs:82-104`** — `SelectAssetCoins` iterates requirements in order and greedily claims coins. A coin carrying multiple assets is locked to the first requirement that encounters it, even if reassigning it would satisfy all requirements.

Example:
- Coin A: [AssetX: 5, AssetY: 10], Coin B: [AssetX: 10]
- Requirements: [AssetX: 10, AssetY: 10]
- Greedy: claims A for X (total=5), claims B for X (total=15) ✓. Then Y: A already reserved → throws `NotEnoughFundsException`
- Optimal: B→X, A→Y → both satisfied

This is a known limitation of greedy allocation but worth a `// TODO` comment so future maintainers know it's intentional, not accidental. For single-asset sends (the 99% case) this is fine.

---

### 🔵 Missing: zero test coverage for `SelectAssetCoins` / asset-aware engine path

The new `SelectAssetCoins` method and the 4-phase engine flow (reserve assets → check if BTC covered → run strategies on remainder → merge) have no tests. All existing engine tests are for `BuildBuckets` only. Suggest adding at minimum:
- Asset coins reserved correctly + BTC contribution subtracted
- Asset-only result when asset coins cover BTC target
- `NotEnoughFundsException` on insufficient asset balance
- Merged result has correct `SelectedCoins`, `TotalValue`, `Change`

---

### Nit: dead parameter

**`BranchAndBoundStrategy.cs:118`** — `PickBetter` accepts `CoinSelectionPolicy policy` but never reads it. Remove or use.

---

**Summary:** The critical bugs are fixed. What's left is two waste-calculation inconsistencies (P2, P3) and test coverage. P2 can cause suboptimal coin selection in edge cases. P4 is a known greedy limitation worth documenting. None of these are blockers if you're comfortable with the trade-offs, but P2 is a one-line fix in two files — worth doing.

⚠️ **Protocol-critical flag:** This code selects VTXOs for spending. Human review required before merge.

PR #126, opened 2026-06-12
feat: arkd signer rotation support

d4rp4t:
Summary

- Adds ServerKeyRotationSweepPolicy — a new sweep policy that automatically re-enrolls VTXOs locked under a deprecated Arkade server signing key into the current one
- The policy uses DeprecatedSigners from ArkServerInfo to identify affected VTXOs; it sweeps coins whose signer key appears in the deprecated list and whose cutoff timestamp hasn't passed yet (or has no cutoff at all)
- Fixes cutoff_date interpretation: the field is a Unix timestamp in seconds, not milliseconds

Motivation

When an Arkade server rotates its signing key, existing VTXOs remain locked to the old key. Without intervention, they expire under the deprecated signer. This policy hooks into the existing SweeperService pipeline and migrates those VTXOs reactively — triggered on each VTXO change — rather than requiring manual user action.


ghost (changes requested):
## Code Review — `feat: arkd signer rotation support` (#126)

Solid concept — the sweep policy is clean, tests are well-structured, and the core filtering logic is correct. But there are a few issues to address, one of which is a **silent correctness bug** that would mask a class of VTXOs from being swept.

---

### 🔴 BUG — Dead null check: `(long?)ds.Value is null` is unreachable

**`ServerKeyRotationSweepPolicy.cs:17`**

```csharp
.Where(ds => ds.Value > now || (long?)ds.Value is null)
```

`DeprecatedSigners` is `Dictionary<ECXOnlyPubKey, long>`. The value type is `long`, not `long?`. Casting a `long` to `long?` always produces a non-null `Nullable<long>` — the `is null` branch is **never true**.

The TODO comment acknowledges this, but the current code is misleading: it reads as if "no cutoff" signers would be swept, when in fact they silently won't be (once the protobuf field becomes optional and the transport layer maps missing values to `0L`, those entries will have `Value == 0` which is `< now`, so they'll be **skipped**).

**Fix:** Until the protobuf is updated to `optional int64`, the default wire value for a missing `cutoff_date` is `0`. You need:
```csharp
.Where(ds => ds.Value == 0 || ds.Value > now)
```
This treats `0` (unset/no cutoff) as "sweep immediately" rather than "expired in 1970". This matches the Rust SDK's behavior where `cutoff_date == 0` means no deadline.

Alternatively, document that `0` is not a valid cutoff and leave the current behavior, but then you're relying on the server never omitting the field — which contradicts the TODO.

---

### 🟡 Missing `[EnumeratorCancellation]` attribute

**`ServerKeyRotationSweepPolicy.cs:10`**

```csharp
public async IAsyncEnumerable<ArkCoin> SweepAsync(IEnumerable<ArkCoin> coins, CancellationToken cancellationToken = default)
```

The existing `SwapSweepPolicy` correctly annotates this parameter with `[EnumeratorCancellation]` (`SwapSweepPolicy.cs:11`). Without it, callers using `await foreach` with a `WithCancellation(token)` won't propagate their token into this method. The `cancellationToken` parameter will always be `default`.

The `SweeperService.ExecutePoliciesAsync` (line 215) doesn't pass a token via `WithCancellation()` today, so this is latent — but it's a correctness gap that'll bite when someone does. Match the pattern from `SwapSweepPolicy`.

---

### 🟡 No DI registration

`ServerKeyRotationSweepPolicy` is not registered in `ServiceCollectionExtensions.AddArkCoreServices()`. The `SweeperService` consumes `IEnumerable<ISweepPolicy>` via DI — without registration, this policy will never execute.

Is the intent for consumers to register it themselves? If so, add an `AddServerKeyRotationSweep()` extension method (similar to `AddArkDelegation()`), or at minimum document it. If it should be on by default, add it to `AddArkCoreServices()`.

---

### 🟢 Minor — missing newline at EOF

**`ServerKeyRotationSweepPolicy.cs:36`** — no trailing newline. Most editors and CI linters flag this.

---

### 🟢 Tests — good coverage, one gap

Test coverage is solid for the happy paths. The one gap is the `cutoff == 0` case (unset protobuf field), which matters because that's the actual wire representation of "no cutoff" before the proto becomes `optional`. Add a test for `cutoff_date = 0` to lock in the expected behavior.

---

### ⚠️ Protocol-critical flag

This PR touches **VTXO lifecycle** — it decides which VTXOs get re-enrolled during signer rotation. A bug here means either: (a) VTXOs under a rotated key silently expire and funds are lost, or (b) VTXOs get unnecessarily swept, burning fees. **Requesting human review.**

ghost (changes requested):
## Code Review — `feat: arkd signer rotation support` (#126)

Solid concept — the sweep policy is clean, tests are well-structured, and the core filtering logic is correct. But there are a few issues to address, one of which is a **silent correctness bug** that would mask a class of VTXOs from being swept.

---

### BUG — Dead null check: `(long?)ds.Value is null` is unreachable

**`ServerKeyRotationSweepPolicy.cs:17`**

```csharp
.Where(ds => ds.Value > now || (long?)ds.Value is null)
```

`DeprecatedSigners` is `Dictionary<ECXOnlyPubKey, long>`. The value type is `long`, not `long?`. Casting a `long` to `long?` always produces a non-null `Nullable<long>` — the `is null` branch is **never true**.

The TODO comment acknowledges this, but the current code is misleading: it reads as if "no cutoff" signers would be swept, when in fact they silently won't be. Once the protobuf field becomes optional and the transport layer maps missing values to `0L`, those entries will have `Value == 0` which is `< now`, so they'll be **skipped** — funds under those signers will expire without being swept.

**Fix:** Until the protobuf is updated to `optional int64`, the default wire value for a missing `cutoff_date` is `0`. You need:
```csharp
.Where(ds => ds.Value == 0 || ds.Value > now)
```
This treats `0` (unset/no cutoff) as "sweep immediately" rather than "expired in 1970". This matches how protobuf3 default values work — a missing `int64` is `0`, not null.

---

### Missing `[EnumeratorCancellation]` attribute

**`ServerKeyRotationSweepPolicy.cs:10`**

```csharp
public async IAsyncEnumerable<ArkCoin> SweepAsync(IEnumerable<ArkCoin> coins, CancellationToken cancellationToken = default)
```

The existing `SwapSweepPolicy` correctly annotates this parameter with `[EnumeratorCancellation]` (`SwapSweepPolicy.cs:11`). Without it, callers using `await foreach` with `.WithCancellation(token)` won't propagate their token into this method — the `cancellationToken` will always be `default`.

The `SweeperService.ExecutePoliciesAsync` (line 215) doesn't pass a token via `WithCancellation()` today, so this is latent — but it's a correctness gap. Match the pattern from `SwapSweepPolicy`:

```csharp
public async IAsyncEnumerable<ArkCoin> SweepAsync(IEnumerable<ArkCoin> coins,
    [EnumeratorCancellation] CancellationToken cancellationToken = default)
```

---

### No DI registration

`ServerKeyRotationSweepPolicy` is not registered anywhere in `ServiceCollectionExtensions.AddArkCoreServices()`. The `SweeperService` consumes `IEnumerable<ISweepPolicy>` via constructor injection — without registration, this policy will never execute at runtime.

Is the intent for consumers to register it themselves? If so, add an `AddServerKeyRotationSweep()` extension method or document the requirement. If it should be on by default, add to `AddArkCoreServices()`:
```csharp
services.AddSingleton<ISweepPolicy, ServerKeyRotationSweepPolicy>();
```

---

### Minor — missing newline at EOF

**`ServerKeyRotationSweepPolicy.cs:36`** — no trailing newline.

---

### Tests — good coverage, one gap

Test coverage is solid. The one gap: add a test for `cutoff_date = 0` (protobuf default for unset `int64`). This is the actual wire representation of "server didn't set a cutoff" and it's the scenario the null-check was trying to handle. Lock in the expected behavior with a test.

---

### Protocol-critical flag

This PR touches **VTXO lifecycle** — it decides which VTXOs get re-enrolled during signer rotation. A bug here means either: (a) VTXOs under a rotated key silently expire and funds are lost, or (b) VTXOs get unnecessarily swept, burning fees. **Requesting human review before merge.**

ghost (changes requested):
## Follow-up Review — Post `0cdf9793`

✅ **DI registration** — Fixed. `AddSingleton<ISweepPolicy, ServerKeyRotationSweepPolicy>()` in `ServiceCollectionExtensions.cs:145` is correct.
✅ **`[EnumeratorCancellation]`** — Was already fixed in `a3351134`.

### 🔴 Still open: Dead null check — `(long?)ds.Value is null` is unreachable

**`ServerKeyRotationSweepPolicy.cs:20`**

```csharp
.Where(ds => ds.Value > now || (long?)ds.Value is null)
```

This was the primary bug from my first review and it's still present. The `(long?)ds.Value is null` branch can never be true — casting a `long` to `long?` always produces a non-null value. The TODO comment acknowledges the proto isn't `optional` yet, but the current code silently drops VTXOs under deprecated signers that have a `cutoff_date` of `0` (the protobuf default for unset `int64`).

**Until the proto is updated**, the safe fix is:
```csharp
.Where(ds => ds.Value == 0 || ds.Value > now)
```

This treats `0` (wire default for missing `cutoff_date`) as "sweep immediately" — matching the Rust SDK behavior. Without this, VTXOs under signers with no cutoff silently expire and **funds are lost**.

The commented-out tests (`cutoff_date = null` scenarios) confirm this case matters. A test for `cutoff_date = 0` should be added alongside the fix.

---

Everything else looks good. Just need this one fix before merge.

⚠️ **Protocol-critical**: This decides which VTXOs get re-enrolled during signer rotation. Human review still required.

ghost (approved):
## Follow-up Review — Post `1b2c564`

All three issues from my previous reviews are resolved:

✅ **Dead null check fixed** (`ServerKeyRotationSweepPolicy.cs:18`) — `ds.Value == 0 || ds.Value > now` correctly treats protobuf default `0` as "sweep immediately" instead of the unreachable `(long?)ds.Value is null`. This prevents VTXOs under no-cutoff deprecated signers from silently expiring.

✅ **Tests uncommented and fixed** — Both `Sweeps_coin_under_deprecated_signer_with_no_cutoff` and `Sweeps_coins_under_all_due_now_deprecated_signers` now use `Dictionary<ECXOnlyPubKey, long>` with `0` values, matching the actual type. Good coverage.

✅ **TODO comments removed** — Clean.

File rename (`SweeperServiceTests.cs` → `Sweeper/SweeperServiceTests.cs`) is organizational only, no logic change.

LGTM. Approving from code review standpoint.

⚠️ **Protocol-critical**: This PR governs which VTXOs get re-enrolled during signer rotation — a bug here means funds silently expire or unnecessary sweep fees. **Human approval still required before merge.**

PR #127, opened 2026-06-12
fix: add ECXOnlyPubKeyComparer

d4rp4t:
# Summary

ECXOnlyPubKey in NBitcoin implements IComparable<T> and overrides GetHashCode (by byte value), but does not implement IEquatable<T> or override Equals/== — so equality falls back to reference identity. Several places in the SDK were silently broken as a result.

- DelegateContractDelegationTransformer — used .Equals() to match a contract's delegate key against the signer's public key; always returned false for distinct instances with the same bytes, causing valid delegates to be rejected
- NofNMultisigTapScript.Parse — used HashSet<ECXOnlyPubKey> to deduplicate owners; reference equality meant duplicate keys in malformed scripts would pass through undetected
- Transport layer — Dictionary<ECXOnlyPubKey, long> for DeprecatedSigners was created without a comparer; key lookups by value would never match

# Fix

Adds ECXOnlyPubKeyComparer — a sealed singleton IEqualityComparer<ECXOnlyPubKey> that compares by the 32-byte x-coordinate and delegates GetHashCode to the existing NBitcoin override. All Dictionary and HashSet constructions keyed on ECXOnlyPubKey now pass ECXOnlyPubKeyComparer.Instance explicitly.

### DIFF
--- a/NArk.Abstractions/Extensions/ECXOnlyPubKeyComparer.cs
+++ b/NArk.Abstractions/Extensions/ECXOnlyPubKeyComparer.cs
@@ -1,0 +1,26 @@
+using NBitcoin.Secp256k1;
+
+namespace NArk.Abstractions.Extensions;
+
+/// <summary>
+/// Value-equality comparer for <see cref="ECXOnlyPubKey"/>.
+/// NBitcoin's <see cref="ECXOnlyPubKey"/> overrides <see cref="object.GetHashCode"/> by byte value
+/// but not <see cref="object.Equals(object)"/>, so <c>Equals</c> falls back to reference equality.
+/// Pass this comparer to any <see cref="System.Collections.Generic.HashSet{T}"/> or
+/// <see cref="System.Collections.Generic.Dictionary{TKey,TValue}"/> keyed on <see cref="ECXOnlyPubKey"/>.
+/// </summary>
+public sealed class ECXOnlyPubKeyComparer : IEqualityComparer<ECXOnlyPubKey>
+{
+    public static readonly ECXOnlyPubKeyComparer Instance = new();
+    private ECXOnlyPubKeyComparer() { }
+
+    public bool Equals(ECXOnlyPubKey? x, ECXOnlyPubKey? y)
+    {
+        if (ReferenceEquals(x, y)) return true;
+        if (x is null || y is null) return false;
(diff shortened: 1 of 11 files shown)


ghost (approved):
## Re-review: `fix: add ECXOnlyPubKeyComparer`

This is a completely different PR from the previous round — force-pushed with new content. Reviewing fresh.

**Root cause analysis: correct.** `ECXOnlyPubKey` overrides `GetHashCode` by byte value but not `Equals`/`==`, so reference equality was silently used everywhere. The three affected sites are all real bugs:
- `DelegateContractDelegationTransformer` — delegate key match always returned false → valid delegates rejected
- `NofNMultisigTapScript.Parse` — `HashSet<ECXOnlyPubKey>` deduplication was broken → malformed scripts with duplicate keys could slip through
- Transport layer `Dictionary<ECXOnlyPubKey, long>` — key lookups by value would never hit

**Implementation: clean.**
- Sealed singleton — correct pattern for a stateless comparer ✅
- `Equals`: null guards, reference short-circuit, then `SequenceEqual` on `ToBytes()` ✅
- `GetHashCode`: delegates to NBitcoin's byte-based override — consistent with `Equals` ✅
- All call sites updated, both Grpc and Rest transports, all test fixtures ✅

**Nits:**

1. **`ToBytes()` allocates on every `Equals` call.** For hot paths (e.g. repeated dictionary lookups over a large `DeprecatedSigners` map) this creates GC pressure. Low priority now, but worth noting. If NBitcoin exposes `WriteToSpan` or similar, a stackalloc comparison would be zero-alloc.

2. **No unit test for `ECXOnlyPubKeyComparer` itself.** The test file changes are all fixture updates. A dedicated test asserting:
   - Two distinct instances with identical bytes → `Equals` returns true
   - `HashSet` deduplication works
   - Null handling
   …would make this PR self-documenting and catch regressions if NBitcoin ever changes its internals.

**Verdict: ✅ Approved.** The bug fix is correct and complete. The two nits are non-blocking.

> ⚠️ Note: key matching in `DelegateContractDelegationTransformer` is contract/signing-adjacent — flagging for human sign-off per protocol review policy, even though the change is clearly correct.

Kukks (approved):
ACK as a fix, but this will bite us again later. Let's add this comparer directly to the type in nbitcoin

PR #129, opened 2026-06-12
test(e2e): close boarding-confirmation flake in settle pipeline tests

Kukks:
## Summary

`UnilateralExitTests.StartExit_IsIdempotentForSameVtxo` (and its siblings sharing `SettleAVtxoAsync`, plus `BoardingTests.CanBoardFromOnchainToVtxo`) failed intermittently with a bare 2-minute `TimeoutException` and zero diagnostics — locally 2 stalls in the first 7 runs.

**Root cause:** the boarding-sync poll declared success as soon as a VTXO row with the funding txid appeared, regardless of confirmation state. The regtest Esplora backend (mempool API) can lag the six just-mined blocks by a few seconds; in that window `BoardingUtxoSyncService` stores the UTXO *unconfirmed* (`ExpiresAt=null`), and `SimpleIntentScheduler` silently filters such coins out (arkd rejects unconfirmed inputs). With the tests' one-shot intent generation cycle (`PollInterval=5h`) nothing ever retried, so the settle stalled until the batch wait timed out.

**Fix:** poll until the boarding VTXO is stored *confirmed* (`ExpiresAt` set). `SyncAsync` re-reads Esplora and upserts on every iteration, so the row flips as soon as the indexer catches up.

**Diagnosability hardening** in the same tests, so the next intermittent failure is attributable from CI output alone:
- fail fast (with the recorded reason) when the intent ends in `BatchFailed`/`Cancelled` instead of timing out blind
- wire timestamped loggers into every settle-pipeline service (intent generation/sync, batch management, boarding sync, scheduler)

A second failure shape was observed once locally (intent registered, round completed server-side per the arkd-wallet broadcast trace, but `BatchSucceeded` never landed client-side). It did not reproduce in 42 instrumented runs; if it recurs, the fail-fast + logging above will surface the terminal state and reason immediately instead of a blind timeout.

## Test plan

- [x] 42 consecutive local green runs of `StartExit_IsIdempotentForSameVtxo` after the fix (was ~2 failures in 7 runs before)
- [x] Full `UnilateralExitTests` file green (4/4)
- [x] `BoardingTests.CanBoardFromOnchainToVtxo` green
- [x] E2E CI on this PR

### DIFF
--- a/NArk.Tests.End2End/BoardingTests.cs
+++ b/NArk.Tests.End2End/BoardingTests.cs
@@ -62,19 +62,27 @@ public class BoardingTests
         var syncService = new BoardingUtxoSyncService(
             contracts, vtxoStorage, clientTransport, utxoProvider);
 
-        // Chopsticks may need a moment to index — retry until the UTXO appears
+        // The Esplora backend may need a moment to index the just-mined blocks —
+        // retry until the UTXO appears *confirmed* (ExpiresAt set). A row synced
+        // while the funding tx still looked unconfirmed has ExpiresAt=null, which
+        // SimpleIntentScheduler silently filters out (arkd rejects unconfirmed
+        // inputs); since the generation cycle below runs only once per
+        // PollInterval, that would stall the whole test. SyncAsync re-reads
+        // Esplora and upserts each iteration, so the row flips to confirmed as
+        // soon as the indexer catches up.
         ArkVtxo? syncedVtxo = null;
         for (var i = 0; i < 10; i++)
         {
             await syncService.SyncAsync();
             var vtxos = await vtxoStorage.GetVtxos();
-            syncedVtxo = vtxos.FirstOrDefault(v => v.TransactionId == fundingTxid);
+            syncedVtxo = vtxos.FirstOrDefault(v => v.TransactionId == fundingTxid && v.ExpiresAt is not null);
             if (syncedVtxo is not null)
                 break;
             await Task.Delay(TimeSpan.FromSeconds(2));
         }
 
-        Assert.That(syncedVtxo, Is.Not.Null, "BoardingUtxoSyncService should find the funded UTXO via Esplora");
+        Assert.That(syncedVtxo, Is.Not.Null,
+            "BoardingUtxoSyncService should find the funded UTXO via Esplora as confirmed (ExpiresAt set)");
         Assert.That(syncedVtxo!.Unrolled, Is.True);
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Arkana Code Review — APPROVE ✅

**Scope**: Test-only (2 E2E test files). No production code changes. No cross-repo impact.

### Root cause fix — correct ✅

The flake root cause is well-diagnosed and the fix is precise:

- `BoardingUtxoSyncService.SyncAsync()` sets `ExpiresAt = null` for unconfirmed UTXOs (confirmed at `BoardingUtxoSyncService.cs:140`)
- `SimpleIntentScheduler` filters `v.Unrolled && v.ExpiresAt is not null` (`SimpleIntentScheduler.cs:43`) — unconfirmed boarding UTXOs are silently skipped
- Old poll predicate matched `TransactionId` only → race: Esplora lags 6 blocks, UTXO synced unconfirmed, scheduler skips it, `PollInterval=5h` means no retry
- New predicate `v.TransactionId == fundingTxid && v.ExpiresAt is not null` waits for confirmed state — correct

Both `BoardingTests.cs` and `UnilateralExitTests.cs` apply the same fix consistently.

### Fail-fast error handling — correct ✅

- `BoardingTests.cs`: `BatchFailed`/`Cancelled` propagates exceptions to all three `TaskCompletionSource` instances (`newIntentTcs`, `newSubmittedIntentTcs`, `newSuccessBatch`) — no blind timeouts
- `UnilateralExitTests.cs`: Same pattern for `newSuccessBatch` — the only TCS in that handler

Exception messages include `IntentTxId`, terminal state, and `CancellationReason` — good diagnostics.

### Logger wiring — correct ✅

All four services (`BoardingUtxoSyncService`, `SimpleIntentScheduler`, `IntentGenerationService`, `IntentSynchronizationService`) accept optional `ILogger?` parameters — verified against source. `LoggerFactory` moved earlier in `SettleAVtxoAsync()` to cover `BoardingUtxoSyncService`. Timestamp format (`HH:mm:ss.fff`) aids correlation with container logs.

### Minor note (non-blocking)

`BoardingTests.cs` still constructs `BoardingUtxoSyncService` without a logger (4-arg ctor). Not a problem — logger is optional and this test already has `Console.WriteLine` tracing — but wiring it would be consistent with `UnilateralExitTests.cs`. Take it or leave it.

LGTM. 42/42 local runs is solid validation.

🤖 Reviewed by Arkana

PR #130, opened 2026-06-12
test(e2e): gate Fulmine liquidity on spendable VTXOs, recover via settle

Kukks:
## Summary

`FulmineLiquidityHelper.EnsureArkLiquidity` gated on Fulmine's raw `/api/v1/balance`, which also counts VTXOs past their renewal deadline. Regtest VTXOs live only **~50 minutes**, so on any long-lived stack Fulmine's balance looks healthy while every offchain send fails with `{"code":2,"message":"missing vtxos"}`. Since `DockerHelper.SendArkdNoteTo` funds nearly every E2E fixture through this helper, the whole suite collapses (observed locally: 34–37 instant `InternalServerError` failures across Asset/BatchSession/Vtxo/Delegation/ChainSwap/SwapManagement fixtures) while the helper keeps declaring liquidity sufficient — it cannot rescue the state it can't see.

**Fix:** gate on the *spendable* subset — sum `/api/v1/vtxos` entries that are not spent, not swept, and ≥1 minute from expiry (falling back to the raw balance if the endpoint is unavailable). Recovery is a **settle**: a settle round re-anchors expired/recoverable funds into a fresh spendable set, so fresh BTC is only boarded when the wallet is genuinely underfunded rather than merely stale.

## Test plan

- [x] Healthy path: Fulmine-funded test (`CanSendAndReceiveBackVtxo`) 5× green; probe logs `spendable=143485754 raw=143485754` and returns immediately
- [x] Recovery mechanics proven on a real stuck faucet: fund boarding + `/api/v1/settle` restored a Fulmine wedged on `missing vtxos` (settle txid returned, balance became spendable), after which the full E2E suite went 50/50 green
- [x] Recovery branch end-to-end against real natural expiry: waited out the ~50-min VTXO lifetime with no manual intervention, then ran the Fulmine-funded test — helper logged `spendable=0 raw=143435754`, settled automatically (txid `3fe06e1e…`), renewed to `spendable=142001396`, test passed
- [x] CI green (Build & Unit Tests + E2E on both runs)

### DIFF
--- a/NArk.Tests.End2End/Common/FulmineLiquidityHelper.cs
+++ b/NArk.Tests.End2End/Common/FulmineLiquidityHelper.cs
@@ -10,31 +10,48 @@ namespace NArk.Tests.End2End.Common;
 public static class FulmineLiquidityHelper
 {
     /// <summary>
-    /// Ensures Fulmine has enough ARK VTXOs by funding its boarding address, mining, and settling.
+    /// Ensures Fulmine has enough *spendable* ARK VTXOs by settling (which renews
+    /// expired/recoverable VTXOs) and, when genuinely underfunded, boarding fresh BTC first.
     /// Call this after Boltz is healthy but before tests that create BTC→ARK or reverse swaps.
     /// </summary>
+    /// <remarks>
+    /// The raw <c>/api/v1/balance</c> amount also counts VTXOs that have passed their
+    /// renewal deadline — Fulmine reports them in the balance but a plain offchain send
+    /// fails with <c>{"code":2,"message":"missing vtxos"}</c>. VTXOs on the regtest stack
+    /// live only ~50 minutes, so a long-lived stack hits this constantly. Gate on the
+    /// spendable subset (from <c>/api/v1/vtxos</c>) instead, and recover via settle:
+    /// a settle round re-anchors the expired funds into a fresh spendable set.
+    /// </remarks>
     public static async Task EnsureArkLiquidity(long minBalance = 200_000, int maxAttempts = 20)
     {
         var fulmineHttp = new HttpClient { BaseAddress = new Uri(SharedSwapInfrastructure.FulmineEndpoint.ToString()) };
 
         for (var attempt = 0; attempt < maxAttempts; attempt++)
         {
-            var arkBalance = await GetFulmineArkBalance(fulmineHttp);
-            Console.WriteLine($"[FulmineLiquidity] ARK balance: {arkBalance} sats (attempt {attempt}, need {minBalance})");
-            if (arkBalance >= minBalance) return;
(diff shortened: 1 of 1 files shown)


ghost (approved):
## ✅ LGTM — approve with notes

**Scope**: test-only (`FulmineLiquidityHelper.cs`). No production code, no public API changes, no cross-repo impact.

**Core logic is sound.** Settle-before-fund is the right fix: expired VTXOs get re-anchored, fresh BTC is only boarded when genuinely underfunded, and the `-1` fallback preserves the old behaviour if `/api/v1/vtxos` is unavailable. Good defensive design.

---

### Findings (non-blocking)

**1. VTXOs without `expiresAt` are silently excluded — `GetFulmineSpendableBalance:166`**

```csharp
if (!long.TryParse(vtxo?["expiresAt"]?.ToString(), out var expiresAt) || expiresAt <= cutoff) continue;
```

If a VTXO has no `expiresAt` field (or it's null/non-numeric), `TryParse` fails → `continue` → VTXO excluded from spendable. This means a VTXO without expiry info is treated as *expired*, which under-counts the balance. In Ark all VTXOs should have expiry, but if Fulmine ever returns one without it (e.g. a boarding VTXO mid-confirmation), the helper would over-settle.

**Suggestion**: treat missing-expiry VTXOs as spendable (they haven't expired):
```csharp
if (long.TryParse(vtxo?["expiresAt"]?.ToString(), out var expiresAt) && expiresAt <= cutoff) continue;
```

**2. Field name fragility — no compile-time contract with Fulmine API**

The method relies on camelCase field names (`isSpent`, `isSwept`, `expiresAt`) parsed from raw JSON. The arkd REST transport in this same repo uses snake_case (`spent_by`, `is_swept`, `expires_at`) — see `RestClientTransport.Vtxo.cs:205-217`. Fulmine's API is separate so this is presumably correct (your local test logs confirm it works), but if Fulmine ever aligns with arkd's naming, the method silently returns -1 and falls back to raw balance.

Not blocking, but consider adding a diagnostic log when `vtxos` array is non-empty yet `spendable == 0` — that would catch a serialization drift early.

**3. `RetryWithSettle` still reports raw balance — `FulmineLiquidityHelper.cs:72`**

The `RetryWithSettle` retry loop still logs `GetFulmineArkBalance` (raw). Cosmetic, but now that you have the spendable probe it'd be useful in retry logs too.

---

### Not flagging for protocol-critical review

This is test infrastructure only — no changes to VTXO handling, signing, forfeit paths, or exit logic in production code. The fallback to raw balance ensures the worst case is the old behavior (unnecessary funding), not missed bugs.

Clean, well-scoped fix for a real problem. Ship it.

PR #128, opened 2026-06-12
feat(transport): advertise target arkd build via X-Build-Version header

d4rp4t:
Every outgoing request to arkd now carries an X-Build-Version: 0.9.7 header, so the server can enforce or log SDK compatibility at the transport layer.

Changes:

- ArkdVersion — single source of truth: TargetBuild = "0.9.7" constant + two InjectHeader extension methods (one on HttpClient, one on Metadata) used by both transports
- BuildVersionInterceptor — gRPC interceptor that copies existing metadata and appends X-Build-Version to every call (unary, server-streaming, client-streaming, duplex)
- GrpcClientTransport / GrpcDelegatorProvider — wired via channel.CreateCallInvoker().Intercept(new BuildVersionInterceptor())
- RestClientTransport — _http.InjectHeader() in both constructors; guard against duplicate headers on repeated calls
- BuildVersionHeaderTests — 4 unit tests covering: constant value, HTTP header injection, idempotency, and RestClientTransport constructor end-to-end

To bump the target version in the future: change ArkdVersion.TargetBuild in one place.

PR #131, opened 2026-06-12
Feat: GetInfo digest

d4rp4t:
Based on #128 , i'll rebase once merged. 

Summary

- Adds X-Digest header derived from the server-info digest returned by GetServerInfoAsync; the digest is stored in a per-transport DigestHolder (volatile string?, no lock — individual reference reads/writes are atomic) and injected by the same interceptor/handler
- DIGEST_MISMATCH response → clears the DigestHolder, throws DigestMismatchException; CachingClientTransport catches it in GetServerInfoAsync and all pass-through methods (Task + IAsyncEnumerable), calls InvalidateServerInfoCache(), then re-throws so the caller can retry


ghost (changes requested):
## Code Review — arkade-os/dotnet-sdk#131 "Feat: GetInfo digest"

Overall: solid design. The `DigestHolder` volatile-ref approach is correct, the `Guard`/`GuardStream` pattern in `CachingClientTransport` is clean, and cross-SDK alignment (ts-sdk, rust-sdk, kotlin proto all have `digest` already) is good. Four issues to fix, two of which are bugs.

---

### 🐛 BUG: REST `RestClientTransport(HttpClient)` constructor missing `BuildVersionHandler`

**`NArk.Core/Transport/RestClient/RestClientTransport.cs`** — second constructor (the one accepting an external `HttpClient`):

```csharp
public RestClientTransport(HttpClient http)
{
    _http = http;
    _baseUri = http.BaseAddress?.ToString().TrimEnd('/') ?? "";
    _http.InjectHeader(); // ← only adds X-Build-Version default header
}
```

The first constructor correctly wires `BuildVersionHandler` into the pipeline:
```csharp
_http = new HttpClient(new BuildVersionHandler(_digestHolder) { InnerHandler = new HttpClientHandler() })
```

But the second constructor does **not**. Any code constructing `RestClientTransport` with an external `HttpClient` gets:
- ❌ No `X-Digest` header on requests (only `BuildVersionHandler.SendAsync` injects it per-request)
- ❌ No `DIGEST_MISMATCH` detection
- ❌ No `BUILD_VERSION_TOO_OLD` detection

The `BuildVersionHeaderTests.RestClientTransport_Constructor_InjectsHeaderOnProvidedHttpClient` test only asserts `X-Build-Version` — it doesn't catch this gap.

**Fix**: Either (a) document that the `HttpClient` constructor is only for testing and the caller is responsible for the handler, or (b) throw if the provided `HttpClient` doesn't already have the handler in its pipeline, or (c) wrap it. Option (a) is probably fine — add a `[Obsolete]` or xml-doc warning. But the test should at minimum verify that `_digestHolder` is wired up somehow.

---

### 🐛 BUG: `GetString` throws on missing `digest` — breaks backward compat with older servers

**`NArk.Core/Transport/RestClient/RestClientTransport.Info.cs`**:

```csharp
var digest = GetString(json, "digest", "digest");
```

`GetString` falls through to `el.GetProperty(snakeCase).GetString()!` when the property is absent — this throws `KeyNotFoundException`. If an older Arkade server doesn't return a `digest` field, the SDK hard-crashes on `GetServerInfoAsync`.

gRPC is fine (protobuf defaults to `""`), but the REST path will blow up.

**Fix**: Use a `GetStringOptional` pattern with `""` as default, or add a `TryGetProperty` guard:
```csharp
var digest = TryGetString(json, "digest", "digest") ?? "";
```

---

### ⚠️ DESIGN: `GrpcDelegatorProvider` has an isolated, never-populated `DigestHolder`

**`NArk.Core/Transport/GrpcClient/GrpcDelegatorProvider.cs`**:

```csharp
private readonly DigestHolder _digestHolder = new();
```

This `DigestHolder` is created but never written to — the delegator has no `GetServerInfoAsync`. The `BuildVersionInterceptor` will read `digestHolder.Digest` on every call and always get `null`, so `X-Digest` is never sent on delegator RPCs.

If the Arkade server validates `X-Digest` on delegator endpoints, every delegator call will fail with `DIGEST_MISMATCH`. If the server ignores missing digests, this is fine but wasteful — consider sharing the transport's `DigestHolder` or removing it from the delegator entirely.

**Question for @d4rp4t**: Is this intentional? Does the delegator service skip digest validation?

---

### 📝 MINOR: Missing EOF newline in `Exceptions.cs`

**`NArk.Core/Exceptions.cs:26`**: File ends without a trailing newline. Trivial but produces noise in future diffs.

---

### ✅ What looks good

- **`DigestHolder` with `volatile`**: Correct. Single-reference reads/writes are atomic on .NET; volatile ensures visibility. No lock needed.
- **`Guard`/`GuardStream` in `CachingClientTransport`**: Clean pattern. `GuardStream` correctly catches `DigestMismatchException` from `MoveNextAsync` during iteration.
- **`BuildVersionInterceptor`**: Covers all four gRPC call types (unary, server-streaming, client-streaming, duplex). `GuardAsync` wrapping is correct.
- **`ArkServerInfo.Digest` placement**: Inserted before optional parameters — positional callers (tests) are all updated. Record is a `record` so `with` expressions still work.
- **Cross-SDK alignment**: ts-sdk, rust-sdk, and kotlin proto all have `digest` on their server-info types already. This brings dotnet-sdk in line.
- **Exception hierarchy**: `DigestMismatchException` and `IncompatibleSdkVersionException` are clean, well-documented, and not caught internally (correct — caller decides retry policy).
- **Test coverage**: `BuildVersionHeaderTests` cover the header injection basics. Existing test factories updated for the new `Digest` parameter.

---

**Verdict**: Request changes for the two bugs (REST constructor + `GetString` crash on older servers). The design question on `GrpcDelegatorProvider` should be answered but isn't blocking. Once the REST constructor and backward compat issues are addressed, this is good to go.

🤖 _Reviewed by Arkana_

ghost (approved):
## Follow-up Review — commit `b7d8cf5` ("add tests and fix issues found by arkana")

All four issues from my previous review are addressed correctly:

### ✅ BUG FIX: REST `digest` backward compat
`RestClientTransport.Info.cs:49` — `TryGetProp` with `""` fallback. Older servers without `digest` no longer crash.

### ✅ BUG FIX / DOC: `RestClientTransport(HttpClient)` constructor
`RestClientTransport.cs:40-55` — Comprehensive XML doc explains the limitation. New test `RestClientTransport_HttpClientConstructor_DoesNotInjectXDigest` documents the gap explicitly.

### ✅ DESIGN: `GrpcDelegatorProvider` DigestHolder
`GrpcDelegatorProvider.cs:16-17` — Field removed, inline `new DigestHolder()` with comment. Clean.

### ✅ MINOR: EOF newline in `Exceptions.cs`
Fixed.

### New test coverage (`DigestTests.cs`) — solid
- `DigestHolderTests`: basic lifecycle
- `BuildVersionHandlerDigestTests`: header injection, DIGEST_MISMATCH throw+clear, case-insensitive matching, BUILD_VERSION_TOO_OLD, non-matching errors pass through
- `CachingTransportDigestMismatchTests`: cache invalidation on GetServerInfo, stale-cache rejection, passthrough invalidation, stream invalidation, re-fetch after recovery

**One minor nit** (non-blocking): `BuildVersionHandlerDigestTests` creates additional `HttpClient` instances in individual tests via `MakeClient()` but `TearDown` only disposes `_client` from `SetUp`. The per-test clients leak. Harmless in tests but easy to fix with `using var client = MakeClient(...)`.

LGTM — approving. Still flagging for human review since this touches protocol-level digest verification that gates VTXO operations.

🤖 Reviewed by Arkana

ghost (approved):
## Follow-up Review — commit `b7d8cf5` ("add tests and fix issues found by arkana")

All four issues from my previous review are addressed correctly:

### ✅ BUG FIX: REST `digest` backward compat
`RestClientTransport.Info.cs:49` — `TryGetProp` with `""` fallback. Older servers without `digest` no longer crash.

### ✅ BUG FIX / DOC: `RestClientTransport(HttpClient)` constructor
`RestClientTransport.cs:40-55` — Comprehensive XML doc explains the limitation. New test `RestClientTransport_HttpClientConstructor_DoesNotInjectXDigest` documents the gap explicitly.

### ✅ DESIGN: `GrpcDelegatorProvider` DigestHolder
`GrpcDelegatorProvider.cs:16-17` — Field removed, inline `new DigestHolder()` with comment. Clean.

### ✅ MINOR: EOF newline in `Exceptions.cs`
Fixed.

### New test coverage (`DigestTests.cs`) — solid
- `DigestHolderTests`: basic lifecycle
- `BuildVersionHandlerDigestTests`: header injection, DIGEST_MISMATCH throw+clear, case-insensitive matching, BUILD_VERSION_TOO_OLD, non-matching errors pass through
- `CachingTransportDigestMismatchTests`: cache invalidation on GetServerInfo, stale-cache rejection, passthrough invalidation, stream invalidation, re-fetch after recovery

**One minor nit** (non-blocking): `BuildVersionHandlerDigestTests` creates additional `HttpClient` instances in individual tests via `MakeClient()` but `TearDown` only disposes `_client` from `SetUp`. The per-test clients leak. Harmless in tests but easy to fix with `using var client = MakeClient(...)`.

LGTM — approving. Still flagging for human review since this touches protocol-level digest verification that gates VTXO operations.

🤖 Reviewed by Arkana

PR #132, opened 2026-06-12
feat: arkd signer-rotation handling (reconciliation, discovery, detection)

Kukks:
## Summary

End-to-end SDK support for **arkd signer-key rotation**. When the operator rotates its signing key, a wallet's existing contracts/addresses remain derived under the old key. This keeps each wallet aligned with the current signer, discovers funds stranded under deprecated signers, and provides the detection signal consumers need.

## Carries #126, #128, #131 (folded in)

This branch is **self-contained** and lands directly on `master`. It carries three now-closed fork PRs whose work it builds on:
- **#128** (`X-Build-Version` headers) — advertises the SDK's target arkd build so the server can reject too-old SDKs.
- **#131** (GetInfo digest — itself includes #128) — `X-Digest`; the `ServerInfoChanged` detection signal is raised off #131's cache invalidation.
- **#126** (`ServerKeyRotationSweepPolicy`) — the before-cutoff sweep that re-enrolls deprecated-signer VTXOs under the current signer.

> **Fix carried over #126:** the policy filtered `coin.SignerDescriptor` (the *user* key) against the deprecated *server* keys, so it never matched a real VTXO — the sweep was a no-op in production. Fixed to match `coin.Contract.Server`, with realistic unit tests + a regression test, plus a full-stack E2E (`ServerKeyRotationSweeperTests`) proving deprecated-signer funds are auto-swept to the current signer. (Worth fixing in #126 upstream too.)

Review fixes + rotation docs from #133 (@d4rp4t) are merged in.

## Net-new in this PR

- **`SingleKeyVtxoRecoveryService`** — SingleKey recovery never probed `DeprecatedSigners` (only HD did), so funds stranded under a rotated signer couldn't be rediscovered. Probes the registered discovery providers once with the flat `tr(pubkey)` descriptor; the indexer provider already cross-products `{current ∪ deprecated signers}`.
- **`EnsureDefaultAsync` + `ISingleKeyDefaultEnsurer`** — fixes the early-return that prevented minting a current-signer default on an already-populated wallet after rotation; idempotently ensures it and returns its script.
- **`ServerInfoChanged` event** (`IServerInfoCacheInvalidation`) — a plain `EventHandler<>` raised from `CachingClientTransport.InvalidateServerInfoCache()` (the single choke-point #131's digest-mismatch path calls), DI-aliased to the same transport singleton. Lets in-process services react to a server-info/signer change without polling.
- **`ContractReconciliationService`** — hosted service (mirrors `SweeperService` lifecycle) that, on `WalletSaved` / `ServerInfoChanged` / a startup pass, ensures each SingleKey wallet's current-signer Default and supersedes stale `Source="Default"` rows.
- **`ServerKeyRotationSweepPolicy` fix + E2E** — see the note above.
- Rotation re-enrollment regression test locking in that swept/recoverable coins re-land under the **current** signer.

## Funds model (3 regimes)

A deprecated-signer VTXO: (1) **before cutoff** → collaboratively swept under the current signer (#126); (2) **after cutoff, before VTXO expiry** → waits (operator refuses the old key); (3) **after expiry** (`IsRecoverable`) → re-enrolls into a batch under the current signer via the existing intent scheduler, with `BatchSession` skipping forfeit for swept inputs so the old key isn't needed. Coin-gathering is keyed on VTXO **script**, not contract Active-state, so deactivating a stale default never strands its funds.

## Tests

`NArk.Tests`: **533 passing**. Plus `NArk.Tests.End2End/ServerKeyRotationSweeperTests` — a full-stack sweep-migration test (runs when the stack advertises a deprecated signer; skips otherwise).


ghost (changes requested):
## 🔍 Aggressive Code Review — arkade-os/dotnet-sdk#132

**Reviewer:** Arkana (automated)
**Verdict:** Request changes — **protocol-critical code (signer rotation, VTXO sweep paths, forfeit-adjacent)** requires human sign-off even though the overall design is sound.

---

### ⚠️ PROTOCOL-CRITICAL — Human Review Required

This PR touches signer-key rotation, VTXO sweep selection, and contract lifecycle. Bugs here directly affect fund recoverability. The three-regime model (before-cutoff sweep, post-cutoff wait, post-expiry re-enroll) is well-designed, but the implementation needs a few eyes on the following:

---

### Findings

#### 🟡 P1 — `ServerKeyRotationSweepPolicy.cs:17-19` — Cutoff boundary semantics

```csharp
var recoverableKeyHexes = serverInfo.DeprecatedSigners
    .Where(ds => ds.Value > now || ds.Value == 0)
    .Select(...)
```

The filter uses strict `>`. If `ds.Value == now` (the exact second the cutoff expires), the coin is **not** swept. This is likely the safe direction (don't sweep at the boundary), but **confirm with arkd** what happens when a client tries to collaboratively spend at `t == cutoff`. If arkd rejects at `>=`, this is correct. If arkd accepts at `==`, there's a one-second window where a sweepable coin is missed (it would be caught on the next sweep cycle, so low severity).

Also: `DateTimeOffset.UtcNow.ToUnixTimeSeconds()` truncates sub-second precision. Verify arkd cutoff timestamps are also in whole seconds to avoid off-by-one across the boundary.

#### 🟢 P2 — `SingleKeyVtxoRecoveryService.cs:97` — `ToEntity(wallet.Id, serverInfo.SignerKey, ...)` parameter is misleading but harmless

```csharp
var entity = contract.ToEntity(wallet.Id, serverInfo.SignerKey, activityState: ContractActivityState.Active)
```

`ToEntity` never reads `defaultServerKey` in its body — it's only used by `GetArkAddress()`. The contract's actual signer key is correctly preserved via `GetContractData()`. No data loss, but the parameter is confusing. Consider passing `null` and adding a comment explaining why the signer key is already embedded in the contract data.

#### 🟡 P3 — `CachingClientTransport.cs:201` — `GuardStream` doesn't propagate `[EnumeratorCancellation]`

```csharp
private async IAsyncEnumerable<T> GuardStream<T>(IAsyncEnumerable<T> source)
{
    var e = source.GetAsyncEnumerator(); // no CancellationToken
```

`GetAsyncEnumerator()` is called without a `CancellationToken`. The primary token is already baked into the inner stream (passed at the call site), so this works for the current usage. However, if a consumer calls `.WithCancellation(ct)` on the returned `IAsyncEnumerable`, that token is silently dropped.

**Suggested fix:**
```csharp
private async IAsyncEnumerable<T> GuardStream<T>(
    IAsyncEnumerable<T> source,
    [EnumeratorCancellation] CancellationToken cancellationToken = default)
{
    var e = source.GetAsyncEnumerator(cancellationToken);
```

And thread the token through from each call site:
```csharp
=> GuardStream(_inner.GetVtxoByScriptsAsSnapshot(scripts, cancellationToken), cancellationToken);
```

#### 🟡 P4 — `ServiceCollectionExtensions.cs:197,224` — Fragile DI cast

```csharp
services.AddSingleton<IServerInfoCacheInvalidation>(sp =>
    (IServerInfoCacheInvalidation)sp.GetRequiredService<IClientTransport>());
```

This assumes `IClientTransport` is always `CachingClientTransport`. If a consumer registers a custom transport (e.g., a test double, a logging wrapper), this throws `InvalidCastException` at runtime with a confusing stack trace. Consider:
- Registering `CachingClientTransport` by its concrete type as well, or
- Using a safe cast with a clear error: `sp.GetRequiredService<IClientTransport>() as IServerInfoCacheInvalidation ?? throw new InvalidOperationException("Transport must implement IServerInfoCacheInvalidation")`

#### 🟢 P5 — `ContractReconciliationService.cs:87` — Linked CTS never disposed

```csharp
var multiToken = CancellationTokenSource.CreateLinkedTokenSource(_shutdownCts.Token, cancellationToken);
```

`multiToken` is never disposed. This is the same pattern as `SweeperService.cs:44`, so it's pre-existing, but it's a resource leak (linked CTS holds a registration on the parent token until disposed). Low priority since it lives for the process lifetime, but worth a `// TODO` at minimum.

#### 🟢 P6 — `ContractReconciliationService.cs` — No job deduplication

Rapid `ServerInfoChanged` events queue multiple `ReconcileAllJob` items. Each one scans all wallets. The unbounded channel won't lose messages but could cause thundering-herd reconciliation. Consider either:
- A flag-based approach (set a `_reconcileAllPending` flag, worker checks and clears it)
- Or draining duplicate `ReconcileAllJob` entries before processing

Low severity — reconciliation is idempotent and non-blocking for funds safety.

#### 🟢 P7 — `CachingClientTransport.cs:109` — `InvalidateServerInfoCache` fires event synchronously

```csharp
ServerInfoChanged?.Invoke(this, eventArgs);
```

This runs subscriber callbacks inline. A slow `ServerInfoChanged` handler (e.g., `ContractReconciliationService.OnServerInfoChanged`) blocks `InvalidateServerInfoCache`, which is called from `Guard()`/`GuardStream()` catch blocks — meaning it blocks the exception propagation path for every pass-through method. Currently safe because `OnServerInfoChanged` just does `TryWrite` (non-blocking), but fragile if future subscribers do work inline.

#### 🟢 P8 — `ArkHostedLifecycle.cs` — `StopAsync` never disposes services

`ArkHostedLifecycle` calls `StartAsync` on `ContractReconciliationService` (and `SweeperService`, etc.) but `StopAsync` is empty. These services implement `IAsyncDisposable` and unsubscribe from events in `DisposeAsync`. They rely on the DI container calling `DisposeAsync` during shutdown. Confirm this is the intended lifecycle — if the DI container disposes in arbitrary order, there could be a window where events fire into a half-disposed service. Pre-existing pattern, not introduced by this PR.

#### 🟢 P9 — `BuildVersionHandler.cs:25-37` — Response body consumed then re-wrapped

```csharp
var body = await response.Content.ReadAsStringAsync(cancellationToken);
// ... checks ...
response.Content = new StringContent(body, Encoding.UTF8, "application/json");
```

On non-success responses, the body is read, checked, then re-wrapped in a new `StringContent`. This works but loses original response headers (Content-Type params, Content-Encoding, etc.). Also forces the entire error body into memory. Fine for typical error responses, but worth a comment explaining the tradeoff.

#### ✅ P10 — Correctness confirmations (things that are RIGHT)

- **EnsureDefaultAsync removing the "skip if contracts exist" guard** — Correct. After rotation, existing contracts use the old signer. `DeriveContract` upserts on `{Script, WalletId}`, so calling it unconditionally is safe and necessary.
- **Reconciliation deactivating stale defaults doesn't strand funds** — Correct per the design. The sweeper gathers coins by VTXO script, not by contract Active state.
- **DigestMismatchException catch ordering** in `GetServerInfoAsync` — Correct. Specific exception before generic `Exception`, and DigestMismatch propagates (doesn't return stale cache).
- **Cross-SDK compatibility** — ts-sdk, go-sdk, and rust-sdk all already support digest headers. Proto `GetInfoResponse.digest` field 19 is consistent. No breaking changes.
- **Test coverage** — Comprehensive. 7 new test files cover sweep policy, reconciliation, recovery, digest handling, build version headers, and the rotation self-output safety property. The `RotationSelfOutputCurrentSignerTests` test is particularly valuable — it locks in the invariant that swept coins re-land under the current signer.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| P1 | 🟡 Medium | `ServerKeyRotationSweepPolicy.cs:17` | Confirm cutoff boundary semantics with arkd |
| P3 | 🟡 Medium | `CachingClientTransport.cs:201` | `GuardStream` drops `[EnumeratorCancellation]` token |
| P4 | 🟡 Medium | `ServiceCollectionExtensions.cs:197` | Fragile `IClientTransport` → `IServerInfoCacheInvalidation` cast |
| P2 | 🟢 Low | `SingleKeyVtxoRecoveryService.cs:97` | Misleading unused `serverInfo.SignerKey` parameter |
| P5 | 🟢 Low | `ContractReconciliationService.cs:87` | Linked CTS leak (pre-existing pattern) |
| P6 | 🟢 Low | `ContractReconciliationService.cs` | No job deduplication for rapid events |
| P7 | 🟢 Low | `CachingClientTransport.cs:109` | Synchronous event firing in exception path |
| P8 | 🟢 Info | `ArkHostedLifecycle.cs` | StopAsync empty; relies on DI disposal order |
| P9 | 🟢 Info | `BuildVersionHandler.cs:25` | Body re-wrap loses original content headers |

**The design is solid.** The three-regime model, the `ServerInfoChanged` event plumbing, and the reconciliation service lifecycle all follow established patterns in this codebase. The test suite is thorough.

**Requesting changes** because this is protocol-critical code and P1/P3/P4 should be addressed or explicitly acknowledged before merge. **Human sign-off required** per protocol review policy.

d4rp4t (changes requested):
Added docs and few fixes here: https://github.com/arkade-os/dotnet-sdk/pull/133

PR #134, opened 2026-06-13
fix(transport): address signer-rotation review findings

d4rp4t:
## Summary

- **fix(grpc):** `BuildVersionInterceptor` — add `GuardedStreamReader<T>` that
  translates `RpcException(DIGEST_MISMATCH / BUILD_VERSION_TOO_OLD)` on every
  `MoveNext` call for server-streaming and duplex-streaming RPCs. Previously only
  `ResponseHeadersAsync` was guarded, so mid-stream digest mismatches propagated
  as raw `RpcException` and never invalidated the server-info cache.

- **fix(transport):** `CachingClientTransport` — defer `ServerInfoChanged` event
  dispatch to after `_lock.Release()` (captured via `postLockEvent` in `finally`).
  Previously the event fired while the lock was held on both the TTL-refresh path
  and the `DigestMismatchException` catch path, allowing a blocking handler to
  deadlock on a re-entrant `GetServerInfoAsync` call.

- **fix(di):** `ServiceCollectionExtensions` — register `CachingClientTransport`
  as the concrete singleton and alias both `IClientTransport` and
  `IServerInfoCacheInvalidation` to it. Previously `IServerInfoCacheInvalidation`
  was wired by casting `IClientTransport`, which throws `InvalidCastException` at
  runtime if the caller replaces `IClientTransport` with a custom transport.

### DIFF
--- a/NArk.Core/Hosting/ServiceCollectionExtensions.cs
+++ b/NArk.Core/Hosting/ServiceCollectionExtensions.cs
@@ -187,15 +187,14 @@ public static class ServiceCollectionExtensions
         // Register the raw gRPC transport
         services.AddSingleton(_ => new GrpcClientTransport(config.ArkUri));
 
-        // Register IClientTransport with caching wrapper as the default
-        services.AddSingleton<IClientTransport>(sp =>
-        {
-            var inner = sp.GetRequiredService<GrpcClientTransport>();
-            var logger = sp.GetService<ILogger<CachingClientTransport>>();
-            return new CachingClientTransport(inner, logger);
-        });
-        services.AddSingleton<IServerInfoCacheInvalidation>(sp =>
-            (IServerInfoCacheInvalidation)sp.GetRequiredService<IClientTransport>());
+        // Register the caching wrapper as the concrete singleton so both IClientTransport and
+        // IServerInfoCacheInvalidation alias the same instance without an unsafe cast.
+        services.AddSingleton<CachingClientTransport>(sp =>
+            new CachingClientTransport(
+                sp.GetRequiredService<GrpcClientTransport>(),
+                sp.GetService<ILogger<CachingClientTransport>>()));
+        services.AddSingleton<IClientTransport>(sp => sp.GetRequiredService<CachingClientTransport>());
+        services.AddSingleton<IServerInfoCacheInvalidation>(sp => sp.GetRequiredService<CachingClientTransport>());
 
         return services;
     }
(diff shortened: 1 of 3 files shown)


ghost (approved):
## Arkana Code Review — PR #134

**Verdict: Approve with notes.** All three fixes are correct. Flagging for mandatory human review since this touches digest/signer-rotation paths (protocol-critical).

---

### 1. DI Registration (`ServiceCollectionExtensions.cs`) ✅

The old code cast `IClientTransport` → `IServerInfoCacheInvalidation`, which blows up if anyone overrides the transport registration. Registering `CachingClientTransport` as the concrete singleton and aliasing both interfaces to it is the textbook fix. Same pattern already used elsewhere in this file (e.g. `SpendingService` at line 136). No behavioral change, just type safety. Applies identically to both `AddArkNetwork` (gRPC) and `AddArkRestTransport` (REST). Clean.

### 2. Deadlock Fix (`CachingClientTransport.cs`) ✅

The core issue: `SemaphoreSlim(1,1)` is not reentrant. Firing `ServerInfoChanged` while holding `_lock` means any handler that calls `GetServerInfoAsync` deadlocks. The fix captures the event args into `postLockEvent` under the lock and fires in `finally` after `_lock.Release()`. This covers both the TTL-refresh path (line ~96 in diff) and the `DigestMismatchException` catch path (line ~108).

**Correctness notes:**
- On the `DigestMismatchException` path, `ClearCacheCore()` runs under the lock (correct — cache state must be consistent), then `throw` re-raises, then `finally` fires the event after lock release. The exception still propagates to the caller. Correct ordering.
- `ClearCacheCore()` is a clean extract of the two assignments. `InvalidateServerInfoCache()` (the public API) still fires events inline — this is fine since it's never called under the lock.
- The `Guard`/`GuardStream` methods (lines 219–245) still call `InvalidateServerInfoCache` directly, which fires inline. Correct — those methods don't hold `_lock`, so no deadlock risk.

### 3. `GuardedStreamReader<T>` (`BuildVersionInterceptor.cs`) ✅

Previously, server-streaming and duplex-streaming RPCs only guarded `ResponseHeadersAsync`. A `DIGEST_MISMATCH` arriving as a trailing status on `MoveNext` would propagate as a raw `RpcException`, bypassing cache invalidation entirely. This is the right fix — wrap the response stream and translate on every `MoveNext`.

The implementation mirrors `GuardAsync<T>` exactly. Applied to both `AsyncServerStreamingCall` (line 39) and `AsyncDuplexStreamingCall` (line 67). `AsyncClientStreamingCall` is unaffected (no response stream iteration). Correct.

**Minor note:** `ex.Status.Detail.Contains(...)` assumes `Detail` is non-null. gRPC C# defaults `Detail` to `""`, so this is safe in practice. But if a future gRPC version or a custom channel breaks that contract, this would throw `NullReferenceException` inside the `when` filter (swallowed, causes the `catch` to not match). Same pattern as the existing `GuardAsync` — not a regression, but worth a `?.Contains(...) == true` hardening pass across all four catch sites someday.

### Missing test coverage ⚠️

No new tests. Specifically:
- **`GuardedStreamReader`**: No test that a mid-stream `DIGEST_MISMATCH` on a server-streaming RPC is translated to `DigestMismatchException`. This is the primary bug being fixed — should have a test.
- **Deferred event dispatch**: The existing `TtlRefresh_with_changed_digest_raises_TtlExpiry_once` test implicitly validates the new code path, but there's no test that a re-entrant `GetServerInfoAsync` from a handler doesn't deadlock (which was the actual bug).
- **DI registration**: No integration test that resolving `IServerInfoCacheInvalidation` returns the same instance as `IClientTransport`.

None of these are blocking — the code is correct. But for protocol-critical infrastructure, tests prevent regressions.

### ⚠️ Protocol-critical — human review required

This PR modifies digest-mismatch handling and signer-rotation event dispatch. A bug here could cause the SDK to silently use a stale signer key after rotation, leading to invalid forfeit signatures. The code looks correct, but per policy this needs a human sign-off.

---

LGTM. @d4rp4t — consider adding a `GuardedStreamReader` test in a follow-up.

PR #135, opened 2026-06-13
feat(safety): disable a wallet sweep destination orphaned by a signer rotation

Kukks:
## Summary

When the Arkade operator rotates its signer, a wallet's configured **sweep destination** (an `ArkAddress`, which encodes a server signer key) can become stale — it points at the now-deprecated former signer, so sweeping there would land funds back under a key the operator no longer co-signs with.

This adds **destination signer-change safety**, entirely in the SDK:

- **`DestinationSafety.IsStale`** — surgical check: a destination is stale iff its `ServerKey` is in `ServerInfo.DeprecatedSigners` (a key that was ours and rotated away). Current-signer and externally-keyed destinations are untouched.
- **`ContractReconciliationService`** (extended) — on startup / `ServerInfoChanged` / `WalletSaved`, for **any** wallet with a destination (SingleKey or HD): flags a stale one in wallet `Metadata` (`destination:pendingConfirmation`) and raises `IDestinationSafetyNotifier.DestinationDisabled` once (on the set transition); clears the flag when the destination is no longer stale.
- **`DefaultWalletProvider`** — skips a flagged destination, so swept funds go to a self-output instead of the stale address. The protection is in the SDK layer and effective regardless of any UI/confirmation flow.
- **`IDestinationSafetyNotifier`** — new event seam, DI-aliased to the reconciliation singleton (mirrors `IServerInfoCacheInvalidation`); consumers subscribe to react (e.g. notify the user).

Re-confirmation: saving the wallet with a destination that is no longer stale fires `WalletSaved` → reconcile clears the flag.

## Tests

TDD throughout; **553 unit tests green**. New coverage: `DestinationSafetyTests` (stale/current/external/null), `DefaultWalletProviderDestinationSkipTests` (flagged → self-output), and reconciliation flag / clear / no-duplicate-event + HD-destination-enqueue tests.

## Docs

New "Destination Safety" section in `docs/articles/signer-rotation.md` + an `IDestinationSafetyNotifier` usage note in the README, per the SDK doc conventions.

### DIFF
--- a/NArk.Abstractions/Wallets/IDestinationSafetyNotifier.cs
+++ b/NArk.Abstractions/Wallets/IDestinationSafetyNotifier.cs
@@ -1,0 +1,20 @@
+namespace NArk.Abstractions.Wallets;
+
+/// <summary>Raised when a wallet's sweep destination is auto-disabled because an Arkade signer rotation made it stale.</summary>
+public sealed class DestinationDisabledEventArgs : EventArgs
+{
+    public required string WalletId { get; init; }
+    /// <summary>The (now stale) destination address string that was disabled.</summary>
+    public required string Destination { get; init; }
+    /// <summary>The deprecated server key (hex) the destination was pinned to.</summary>
+    public required string DeprecatedServerKey { get; init; }
+}
+
+/// <summary>
+/// Notifies consumers (e.g. a BTCPay plugin) that a wallet's sweep destination was disabled pending
+/// re-confirmation after an Arkade signer rotation. DI-aliased to the same singleton that performs detection.
+/// </summary>
+public interface IDestinationSafetyNotifier
+{
+    event EventHandler<DestinationDisabledEventArgs>? DestinationDisabled;
+}
--- a/NArk.Core/Hosting/ServiceCollectionExtensions.cs
+++ b/NArk.Core/Hosting/ServiceCollectionExtensions.cs
@@ -143,6 +143,7 @@ public static class ServiceCollectionExtensions
         services.AddSingleton<IOnchainService, OnchainService>();
         services.AddSingleton<SweeperService>();
         services.AddSingleton<ContractReconciliationService>();
+        services.AddSingleton<IDestinationSafetyNotifier>(sp => sp.GetRequiredService<ContractReconciliationService>());
         services.AddSingleton<ISweepPolicy, ServerKeyRotationSweepPolicy>();
         services.AddSingleton<PendingArkTransactionRecoveryService>();
         services.AddSingleton<IFeeEstimator, DefaultFeeEstimator>();
--- a/NArk.Core/Services/ContractReconciliationService.cs
+++ b/NArk.Core/Services/ContractReconciliationService.cs
@@ -1,5 +1,6 @@
 using System.Threading.Channels;
 using Microsoft.Extensions.Logging;
+using NArk.Abstractions;
 using NArk.Abstractions.Contracts;
 using NArk.Abstractions.Wallets;
 using NArk.Core.Recovery;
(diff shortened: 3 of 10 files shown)


ghost (commented):
## 🔍 Arkana Code Review — Protocol-Critical

**Verdict: Code is correct and well-structured, but this is protocol-critical (controls where swept funds land). Requires human sign-off before merge.**

---

### Architecture

The design is clean: a static `DestinationSafety.IsStale` predicate, a metadata flag lifecycle managed by the existing `ContractReconciliationService`, and a gating check in `DefaultWalletProvider.ShouldUseDestination`. The event-seam via `IDestinationSafetyNotifier` is well-scoped. DI alias to the existing singleton avoids extra registrations.

### Correctness ✅

- **`DestinationSafety.IsStale`** (`DestinationSafety.cs:20-23`): Correctly identifies stale destinations — only keys present in `DeprecatedSigners` match. External keys and the current signer are excluded by construction. The linear scan over deprecated keys is fine; this set will always be small.

- **Flag lifecycle** (`ContractReconciliationService.cs:210-242`): Set transition fires the event once; already-flagged is a no-op (no duplicate event, no redundant write); clear path passes `null` to `SetMetadataValue` which removes the key. Correct.

- **Sweep gating** (`DefaultWalletProvider.cs:148-151`): `ShouldUseDestination` correctly short-circuits on missing destination or presence of the flag. Since `GetAddressProviderAsync` constructs a fresh provider per call (no caching), the flag is checked at sweep time. Correct.

- **`ReconcileAllAsync` filter change** (`ContractReconciliationService.cs:178`): Removing the `SingleKey` filter from the loop is correct — HD wallets need destination safety too. The `SingleKey` guard is preserved downstream (`ReconcileWalletAsync` line after destination check) so HD wallets only get destination-checked, not contract-reconciled.

- **`OnWalletSaved` broadening** (`ContractReconciliationService.cs:279-281`): Now enqueues HD wallets *with a destination* — enables flag-clearing on re-save. SingleKey wallets still always enqueue. Correct.

### Minor Observations (non-blocking)

1. **Small race window** between `ServerInfoChanged` firing and the reconciliation worker processing the flag. During this window, `GetAddressProviderAsync` could construct a provider with the stale destination. Practically negligible (sub-second, and a sweep still requires a full round to complete), but worth documenting.

2. **Redundant DB load in `ReconcileAllAsync`** (`ContractReconciliationService.cs:183`): `LoadAllWallets` fetches all wallets, but `ReconcileWalletAsync` re-loads each by ID. For HD wallets without destinations, this is a wasted round trip. Minor perf — not a bug.

3. **`DestinationSafety.IsStale` could use `ContainsKey`** instead of iterating (`DestinationSafety.cs:21-22`): Since `DeprecatedSigners` is a `Dictionary<ECXOnlyPubKey, long>` constructed with `ECXOnlyPubKeyComparer.Instance`, `serverInfo.DeprecatedSigners.ContainsKey(destination.ServerKey)` would be O(1) vs O(n). Only matters if you'd ever need to verify `GetHashCode` consistency on `ECXOnlyPubKeyComparer` — the current `obj.GetHashCode()` delegates to NBitcoin's implementation which may or may not be byte-derived. The linear scan is the safe choice if there's any doubt.

4. **Untested path**: `ArkAddress.TryParse` failure on a malformed `wallet.Destination` silently skips the destination check (`ContractReconciliationService.cs:210`). Correct behavior, but no test covers it.

### Test Coverage ✅

Good coverage:
- `DestinationSafetyTests`: stale/current/external/null — all branches of `IsStale`
- `DefaultWalletProviderDestinationSkipTests`: flagged/unflagged/empty/other-metadata — all branches of `ShouldUseDestination`
- `ContractReconciliationServiceTests`: flag-set, flag-clear, no-duplicate-event, HD-wallet-enqueue

### Security

No key exposure, no injection vectors, no unsafe deserialization. The metadata flag is a simple string marker — no auth bypass possible. The `DeprecatedServerKey` hex in the event args is derived from the public key, not secret material.

### Cross-Repo Impact

None. `IDestinationSafetyNotifier` is a new interface with no external consumers. `DefaultWalletProvider` and `ContractReconciliationService` are internal to dotnet-sdk. No other SDK (ts/go/rust) is affected.

---

### ⚠️ Protocol-Critical Flag

This PR controls **where swept funds are routed**. A bug in the stale-detection or flag-gating logic could either:
- Send funds to a rotated-away signer (loss of co-signing capability)
- Block legitimate sweeps to valid destinations (denial of service)

The code looks correct, but per policy this requires **explicit human sign-off** before merge.

cc @Kukks — clean work. The TDD approach and event-seam design are solid.

PR #136, opened 2026-06-13
fix(intents): IntentSynchronizationService tolerates DI-scope disposal race at teardown

Kukks:
## Problem

At host teardown, `IntentSynchronizationService.DisposeAsync` cancels the shutdown token and awaits the running submit loop. The loop's in-flight `intentStorage.GetIntents(...)` (an EF query) can throw `ObjectDisposedException` when the DI `ServiceProvider`/scope is concurrently disposed — instead of honouring cancellation. The loop and `DisposeAsync` only caught `OperationCanceledException`, so the `ObjectDisposedException` escaped and **failed the whole host disposal**, intermittently reddening E2E tests (whichever test loses the timing race at teardown — observed on `FullRecovery_RestoresContracts_Index_AndFunds`, `DeprecatedSignerVtxo_IsAutomaticallySweptToCurrentSigner`, …).

This is a long-standing flake on `master`, surfaced more often by the rc.1 + deprecated-signer E2E baseline.

## Fix

- `DoIntentSubmitLoop` now also catches `ObjectDisposedException` **when the token is cancelled** (shutdown) — stopping the loop quietly. Outside shutdown an `ObjectDisposedException` is a real fault and still propagates (the `when` guard).
- `DisposeAsync` catches `ObjectDisposedException` as a belt-and-suspenders (it only runs during shutdown, so it is always benign there).

## Test

New `IntentSynchronizationServiceTests` deterministically reproduces the race: a mocked in-flight `GetIntents` throws `ObjectDisposedException` during disposal; the test asserts `DisposeAsync` completes without throwing. RED before the fix (the exception escaped `DisposeAsync` via `DoIntentSubmitLoop` → `await _intentSubmitLoop`), green after. Full unit suite **542 green**.

### DIFF
--- a/NArk.Core/Services/IntentSynchronizationService.cs
+++ b/NArk.Core/Services/IntentSynchronizationService.cs
@@ -94,6 +94,14 @@ public class IntentSynchronizationService(
         {
             logger?.LogDebug("Intent submit loop cancelled");
         }
+        catch (ObjectDisposedException) when (token.IsCancellationRequested)
+        {
+            // Shutdown race: the DI scope was disposed (host teardown) while a storage query was in
+            // flight, so an EF/DbContext access threw ObjectDisposedException instead of honouring the
+            // cancellation. Benign during shutdown — stop the loop quietly. Outside shutdown an
+            // ObjectDisposedException is a real fault and still propagates (the `when` guard).
+            logger?.LogDebug("Intent submit loop stopped: service provider disposed during shutdown");
+        }
     }
 
     private async Task SubmitIntent(ArkIntent intentToSubmit, CancellationToken token)
@@ -207,6 +215,12 @@ public class IntentSynchronizationService(
         catch (OperationCanceledException)
         {
             logger?.LogDebug("Intent submit loop cancelled during disposal");
+        }
+        catch (ObjectDisposedException)
+        {
+            // Belt-and-suspenders: an in-flight storage query lost the race with DI-scope disposal at
+            // teardown. Always benign here (DisposeAsync only runs during shutdown).
+            logger?.LogDebug("Intent submit loop observed a disposed service provider during disposal");
         }
 
         logger?.LogInformation("Intent synchronization service disposed");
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Review: APPROVE ✅

Clean, well-scoped fix for a legitimate shutdown race. Not protocol-critical — this is infrastructure/lifecycle code, no VTXO, signing, or round-lifecycle paths are touched.

### Analysis

**The bug:** `DoIntentSubmitLoop` and `DisposeAsync` only caught `OperationCanceledException`. During host teardown, the DI `ServiceProvider` can be disposed concurrently with an in-flight EF query, which throws `ObjectDisposedException` instead of honoring the `CancellationToken`. This escapes both catch blocks and fails `DisposeAsync`, flaking E2E tests.

**The fix covers both timing scenarios:**

1. **Token cancelled before ODE thrown** → `DoIntentSubmitLoop` catches `ObjectDisposedException` via `when (token.IsCancellationRequested)` guard → loop exits cleanly → `DisposeAsync` awaits a completed task.
2. **ODE thrown before token cancelled** → `when` guard is false, exception propagates → `_intentSubmitLoop` faults → `DisposeAsync` catches `ObjectDisposedException` in its own belt-and-suspenders block.

Both paths terminate cleanly. The `when` guard is the correct pattern — it prevents swallowing a real `ObjectDisposedException` during normal (non-shutdown) operation. The unconditional catch in `DisposeAsync` is safe because `DisposeAsync` only executes during shutdown.

### Test

The test (`DisposeAsync_does_not_throw_when_in_flight_query_races_scope_disposal`) deterministically reproduces timing scenario #2 above: the mock throws ODE before `DisposeAsync` cancels the token. This exercises the `DisposeAsync` catch block. The `when` guard path (scenario #1) is not directly tested, but the guard is trivially correct and the overall fix is validated.

One subtlety worth noting: the test calls `shutdownRequestedTcs.TrySetResult()` before `DisposeAsync()`, so depending on task scheduling, the `when` guard *might* see `IsCancellationRequested == false` when it evaluates. In practice this is fine — both code paths converge to the same correct behavior.

### Nits (pre-existing, not blocking)

- `StartAsync` line 44: `CreateLinkedTokenSource(...)` result is assigned to a local `multiToken` that is never disposed. Minor `CancellationTokenSource` leak. Not this PR's problem.

LGTM. Ship it.

---
🤖 Reviewed by Arkana

PR #142, opened 2026-06-17
feat: Arkade delegator server — VTXO-refresh delegation intake

Kukks:
## What

Lets the NArk SDK **be** a delegator (the server side of `fulmine.v1.DelegatorService`), not just consume one. A new `NArk.Delegator` ASP.NET library accepts VTXO-refresh delegations, validates and co-signs them, and persists them — reusing the existing intent storage and pipeline.

## Changes

- **`NArk.Delegator`** (new ASP.NET library): `DelegatorGrpcService` implementing the generated `DelegatorService.DelegatorServiceBase`, exposed over **gRPC + REST** (JSON transcoding off the proto's `google.api.http` annotations). `AddNArkDelegator(...)` / `MapNArkDelegator()` DI + endpoint helpers. `delegator.proto` flipped to `GrpcServices="Both"` in `NArk.Core` (no duplicate `Fulmine.V1` types).
- **`DelegateeService`** (`NArk.Core`): `GetInfoAsync` (advertise pubkey/fee/address) and `AcceptAsync` (intake): parse → reconstruct the delegate contract from the forfeit leaf and **verify it against the VTXO scriptPubKey** → **co-sign the forfeit's delegate path** with the delegator's signer → `reject_replace`/supersede → optional fee check → persist as a `WaitingToSubmit` `ArkIntent` (`ValidFrom = expiry − threshold`) plus the co-signed forfeits.
- `ArkIntent` gains a `PartialForfeits` field (round-tripped through EF Core storage); no new entity/migration.
- **Client (offline-refresh) fixes**, confirmed against fulmine's Go source:
  - declares the **delegator** as the intent cosigner (the offline owner can't cosign the refreshed tree);
  - builds the forfeit as the real Arkade structure — TRUC (v3) `VTXO → forfeit-address(amount+Dust) + P2A anchor`, `ANYONECANPAY|ALL` — instead of a placeholder OP_RETURN.

## Tests

- **Unit:** `dotnet test NArk.Tests` → 559/559 pass.
- **E2E (regtest, `--profile boltz`):** `DelegatorServerTests` (`GetDelegatorInfo` over gRPC + REST against an in-process host) and `DelegatorIntakeTests` (a real client delegation is validated, co-signed, and persisted) — both green against live arkd.

## Scope / follow-up

Ships the delegation **intake** (accept → validate → co-sign → persist), end-to-end verified. The automated **batch refresh** — registering the held intent at `ValidFrom` and joining a batch to submit the stored forfeit (with the operator connector appended) — is the tracked next step; the confirmed forfeit protocol is documented in `docs/articles/delegator-server.md`.

## Docs

README "Running an Arkade Delegator" section + `docs/articles/delegator-server.md` (+ toc). DocFX build green.

### DIFF
--- a/NArk.Abstractions/Intents/ArkIntent.cs
+++ b/NArk.Abstractions/Intents/ArkIntent.cs
@@ -22,6 +22,13 @@ public record ArkIntent(
     string SignerDescriptor
 )
 {
+    /// <summary>
+    /// Partially-signed forfeit transactions (base64), one per input VTXO. Populated for delegated
+    /// intents (the client's pre-signed forfeit co-signed by the delegator); empty for owner-generated
+    /// intents, whose forfeits are constructed at batch time.
+    /// </summary>
+    public string[] PartialForfeits { get; init; } = [];
+
     private sealed class IntentTxIdEqualityComparer : IEqualityComparer<ArkIntent>
     {
         public bool Equals(ArkIntent? x, ArkIntent? y)
--- a/NArk.Core/Models/Options/DelegatorOptions.cs
+++ b/NArk.Core/Models/Options/DelegatorOptions.cs
@@ -1,0 +1,30 @@
+using NBitcoin.Scripting;
+
+namespace NArk.Core.Models.Options;
+
+/// <summary>
+/// Configuration for the Arkade delegator (the server side of the VTXO-refresh delegation service).
+/// </summary>
+public class DelegatorOptions
+{
+    /// <summary>Wallet identifier whose signer the delegator co-signs and joins batches with.</summary>
+    public required string WalletId { get; set; }
+
+    /// <summary>
+    /// The delegator's signing descriptor. Its key appears in the delegate leaf of each delegate
+    /// contract, and is what clients embed after calling GetDelegatorInfo.
+    /// </summary>
+    public required OutputDescriptor DelegateDescriptor { get; set; }
+
+    /// <summary>Service fee advertised via GetDelegatorInfo, as a plain-text amount string (sats).</summary>
+    public string Fee { get; set; } = "0";
+
+    /// <summary>The Arkade address the service fee is paid to. Empty when no fee is charged.</summary>
+    public string DelegatorAddress { get; set; } = "";
+
+    /// <summary>How long before a VTXO's expiry the refresh batch should fire (time-based).</summary>
+    public TimeSpan RefreshThreshold { get; set; } = TimeSpan.FromHours(2);
+
+    /// <summary>How many blocks before a VTXO's height-expiry the refresh should fire.</summary>
+    public uint RefreshThresholdHeight { get; set; } = 144;
+}
--- a/NArk.Core/NArk.Core.csproj
+++ b/NArk.Core/NArk.Core.csproj
@@ -30,7 +30,10 @@
     </ItemGroup>
 
     <ItemGroup>
-        <Protobuf Include="Transport\GrpcClient\Protos\**\*.proto" GrpcServices="Client" ProtoRoot="Transport\GrpcClient\Protos" AdditionalImportDirs="$(PkgGoogle_Api_CommonProtos)/content/protos" />
+        <!-- Client stubs for every proto EXCEPT the fulmine delegator service, which we also implement server-side. -->
+        <Protobuf Include="Transport\GrpcClient\Protos\**\*.proto" Exclude="Transport\GrpcClient\Protos\fulmine\v1\delegator.proto" GrpcServices="Client" ProtoRoot="Transport\GrpcClient\Protos" AdditionalImportDirs="$(PkgGoogle_Api_CommonProtos)/content/protos" />
+        <!-- Delegator service: generate BOTH client and server so NArk.Delegator can implement the server base. -->
+        <Protobuf Include="Transport\GrpcClient\Protos\fulmine\v1\delegator.proto" GrpcServices="Both" ProtoRoot="Transport\GrpcClient\Protos" AdditionalImportDirs="$(PkgGoogle_Api_CommonProtos)/content/protos" />
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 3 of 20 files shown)


ghost (changes requested):
## Arkana Code Review — PR #142: Arkade Delegator Server

**Reviewer:** Arkana (aggressive protocol review)
**Verdict:** Request changes — protocol-critical code, requires human sign-off regardless

Overall this is well-structured work. The contract reconstruction + parity check is a sound design. The forfeit protocol (TRUC v3, ACP|ALL, connector grafted at batch time) matches the Arkade spec. The client-side cosigner fix (delegate instead of owner) is correct and necessary. Several issues below need attention before merge.

---

### 🔴 P0 — Protocol-Critical

**1. Superseding `BatchInProgress` intents is dangerous**
`DelegateeService.cs:193-208` — The overlap query includes `ArkIntentState.BatchInProgress`:
```csharp
states: [ArkIntentState.WaitingToSubmit, ArkIntentState.WaitingForBatch, ArkIntentState.BatchInProgress],
```
When the batch-refresh step is implemented, cancelling a `BatchInProgress` intent while the operator is mid-round could leave the operator holding a partial tree that never completes, or worse, the VTXO could be double-committed (old batch completes on-chain + new delegation persisted for the same outpoints). Today this is safe because batch-refresh isn't implemented yet, but this is a landmine.

**Recommendation:** Either (a) exclude `BatchInProgress` from the supersede query and reject the delegation if any input overlaps a `BatchInProgress` intent, or (b) add a loud `// TODO: PROTOCOL-CRITICAL` comment explaining why this is safe today and what must change when batch-refresh lands.

**2. No server key validation in contract reconstruction**
`DelegateeService.cs:238-241` — The code validates that the delegate key is present in the leaf script but does **not** validate that the server key is present:
```csharp
if (!keys.Any(k => comparer.Equals(k, delegateXOnly)))
    throw new DelegationRejectedException("...");
// ← no check for serverXOnly
```
If a malicious client crafts a script where the third key is NOT the Arkade server key, the reconstructed contract would use `serverInfo.SignerKey` anyway, producing a different scriptPubKey — caught by the parity check at line 174. So this fails safely. But adding an explicit server key check would produce a clearer error message and defend against future refactors that might weaken the parity check.

**3. CLTV-gated delegate contracts are silently rejected**
`DelegateeService.cs:246-248` — The reconstruction always builds a non-CLTV contract (`cltvLocktime` defaults to `null`). If a client's VTXO uses a CLTV-gated delegate contract, the parity check fails with a misleading error ("does not reference a delegate contract bearing this delegator's key"). This is safe but should either:
- Parse the CLTV locktime from the leaf script and pass it to the constructor, OR
- Explicitly document that CLTV contracts are not supported for delegation intake

---

### 🟡 P1 — Security / Correctness

**4. No VTXO existence/unspent validation**
`DelegateeService.cs:139-220` — `AcceptAsync` validates the contract structure and co-signs the forfeit, but never checks whether the referenced VTXOs actually exist on the Arkade server or are unspent. `ComputeRefreshWindowAsync` (line 290) calls `GetVtxosByOutpoints` but only to compute the expiry window — if the VTXO isn't found, it silently falls back to `ValidFrom = UtcNow`. A non-existent or already-spent VTXO would be accepted, co-signed, and persisted.

**Recommendation:** After `ComputeRefreshWindowAsync`, verify that all outpoints were resolved. Reject if any VTXO is missing or already spent.

**5. `string[]` EF Core mapping on SQLite**
`EfCoreIntentStorage.cs` + `ArkIntentEntity.cs` — `PartialForfeits` is `string[]` with no explicit column configuration. PostgreSQL's Npgsql maps this to `text[]` natively, but SQLite has no array type. EF Core will either fail at runtime or silently serialize it in an undocumented way. The repo explicitly supports both providers.

**Recommendation:** Add a JSON value converter for `PartialForfeits` in `ArkIntentEntity.Configure()`, or document that SQLite is unsupported for delegator storage.

**6. Fee validation integer overflow edge case**
`DelegateeService.cs:279-283`:
```csharp
var paid = (ulong)intentPsbt.GetGlobalTransaction().Outputs
    .Where(o => o.ScriptPubKey == feeScript)
    .Sum(o => o.Value.Satoshi);
```
`.Sum()` returns `long`. The cast to `ulong` is safe for valid Bitcoin amounts but `Sum` can throw `OverflowException` on overflow in checked context. Minor, but worth noting.

---

### 🟢 P2 — Improvements

**7. Unit test coverage gap**
`DelegateeServiceTests.cs` — Only `GetInfoAsync` is unit-tested. `AcceptAsync` — the protocol-critical path that validates contracts, co-signs forfeits, and persists intents — has zero unit test coverage. The E2E test covers the happy path but not edge cases (invalid PSBT, wrong delegate key, missing leaf script, fee underpayment, etc.).

**8. `ReadSpendingLeafScript` relies on `Unknown` bag**
`DelegateeService.cs:251-259` — PSBT key type `0x15` (TAP_LEAF_SCRIPT) is read from `input.Unknown`. If a future NBitcoin version natively parses this key type, it would no longer appear in `Unknown` and this code would silently return `null`. Add a comment noting this dependency.

**9. `DelegateeService` registered as Singleton**
`ServiceCollectionExtensions.cs:22` — `AddSingleton<DelegateeService>()` means all requests share one instance. This is fine today (no mutable state except via injected services), but could become a problem if mutable caching is added later. Consider `AddScoped` or at minimum add a comment.

---

### ✅ What's Good

- Contract reconstruction + parity check against scriptPubKey is the right approach — deterministic, doesn't trust the client's claim about which contract they're using
- TRUC (v3) forfeit with proper outputs (forfeit address + P2A anchor) matches the Arkade wire format
- ACP|ALL sighash is correct — allows connector grafting without invalidating the VTXO input signature
- Cosigner fix (delegate key, not owner key) is necessary and correct — the owner is offline at batch time
- Reject/replace + supersede logic is well-structured (modulo the BatchInProgress concern above)
- E2E tests against live regtest arkd give real confidence
- Documentation is thorough

---

⚠️ **This PR touches protocol-critical code paths** (VTXO forfeit signing, delegate contract verification, intent lifecycle). Per project policy: **human review required before merge**, even if all code issues above are addressed.

🤖 _Reviewed by Arkana_

PR #143, opened 2026-06-19
fix: commit hash version field

jackstar12:
This field doesn't exist in the API

### DIFF
--- a/NArk.Swaps/Boltz/Models/Info/VersionResponse.cs
+++ b/NArk.Swaps/Boltz/Models/Info/VersionResponse.cs
@@ -6,7 +6,4 @@ public class VersionResponse
 {
     [JsonPropertyName("version")]
     public required string Version { get; set; }
-
-    [JsonPropertyName("commitHash")]
-    public required string CommitHash { get; set; }
 }


ghost (approved):
## Review: fix: commit hash version field

**Verdict: Approve** ✅

### What this does
Changes `VersionResponse.CommitHash` from `required string` → `string?` and adds an `[Obsolete]` attribute, because the Boltz `/v2/version` endpoint does not return a `commitHash` field.

### Analysis

**Deserialization fix — correct.**  
`System.Text.Json` with `required` will throw if the property is missing from the JSON payload. Since Boltz doesn't return `commitHash`, `GetVersionAsync()` in `BoltzClient.Info.cs:16` would fail on every call. Making it `string?` fixes that.

**No downstream consumers.**  
Searched the entire repo and cross-repo (`/root/arkana/repos`). `CommitHash` is never read — only defined in `VersionResponse.cs` and deserialized in `BoltzClient.Info.cs`. No breakage risk.

**`[Obsolete]` attribute — good.**  
Warns future consumers not to rely on this field. Clear message text.

**Not protocol-critical.**  
This is a Boltz swap info endpoint (version check). No VTXO, signing, forfeit, round lifecycle, or exit path code is touched.

### Nits (non-blocking)
- Trailing blank line before closing `}` at EOF — cosmetic, not worth a round-trip.

LGTM. Clean, minimal, correct fix.

d4rp4t (approved):

Kukks (changes requested):
Just remove it entirely.

Kukks (approved):

